PR #18, opened 2018-11-06
Add graph class

pfistfl:
Add's a Graph Class and some unit tests for it. 
I have worked around the issue of whether we require GraphNode's / PipeOp's for now


berndbischl:
i iwll merge these quickly now, if they seem to move us forward. we can open issues later

PR #20, opened 2018-11-07
Dev graphnode

zzawadz:
Separate GraphNode and PipeOps


pfistfl:
I will merge for now, so I can work on it.

PR #25, opened 2018-11-09
Feature transform

pfistfl:
Introduces a new **PipeOp**: PipeOpFeatureTransform which maps from
[data.table] -> [data.table]

Example: (see train_dt)
```
PipeOpPCA = R6Class("PipeOpPCA",
  inherit = PipeOpFeatureTransform,
  public = list(
    initialize = function(id = "pca") {
      super$initialize(id)
    },
    train_dt = function(dt) {
      pcr = prcomp(as.matrix(dt), center = FALSE, scale = FALSE)
      private$.params = pcr$rotation
      as.data.table(pcr$x)
    },
    predict2 = function() {
      assert_list(self$inputs, len = 1L, type = "Task")
      # FIXME: Make sure dimensions fit (if input did not have full rank)
    }
  )
)
```


berndbischl:
you can always merge in cases like these, for now

Issue #41, opened 2018-11-16
we need a good printer for graphs


berndbischl:
1) the graphical plotter should plot the graph as is

2) the printer needs to show a compact form on the console.
we could implement this by topological sorting

with that i mean we sort the graph in layer. then show it like this:

[ids, 1] >> [ids, 3] >> [ids, 2]

the number is the number of elements in the layer, ids is a concatened string of the ids

zzawadz:
Sth like this?

```
# Pipeline Graph
#  [(OpNULL), 1] >> [(scaler,pca), 2] >>
#  [(featureunion), 1] >> [(classif.rpart), 1] 
```

pfistfl:
I like that.
The other option would be to make the printer have multiple lines which is a little tedious and not super important IMHO.

berndbischl:
we need to figure out whther "layers" really exist in toposorting

berndbischl:
how about we simply print the edge datatable, or its head? if the user wants to see the real graph, he plots it with igraph.

mb706:
Currently the printer does this:
```r
> gr = PipeOpScale$new() %>>% PipeOpPCA$new() %>>% PipeOpLearner$new(mlr_learners$get("classif.rpart"))
> gr
Graph with 3 PipeOps:
             ID         State
1         scale <<UNTRAINED>>
2           pca <<UNTRAINED>>
3 classif.rpart <<UNTRAINED>>
> gr$train(mlr_tasks$get("iris"))
INFO [mlr3] Training learner 'classif.rpart' on task 'iris' ...
[[1]]
NULL

> gr
Graph with 3 PipeOps:
             ID        State
1         scale       <list>
2           pca     <prcomp>
3 classif.rpart <Experiment>
```

PipeOps are sorted topologically, but it is not limited to the head. Thoughts on this?

berndbischl:
ok what you posted is something else, not the edges. but i still like it. check a bit, then colse

berndbischl:
maybe we can add the predeccs and successors nodes as character cols?

mb706:
Still needs predecessors & successors

Issue #36, opened 2018-11-16
traverseGraph should potentially be a method of Graph. it also should be called "map"


berndbischl:

zzawadz:
https://github.com/mlr-org/mlr3pipelines/commit/e5b5bdeb1d827b4bbc0434b2edc1201df0c283d6 - new map function.

berndbischl:
thx. but pleeeeease no simplify = TRUE. at least not as the default.

berndbischl:
obsolete

Issue #33, opened 2018-11-16
remove all instances of "<-"


berndbischl:

zzawadz:
I will be in my today's pull request.

PR #49, opened 2018-11-19
Dev zz

zzawadz:
#33 
#41 - first attempt
#36


R/PipeOpFeatureUnion.R
```diff
@@ -11,23 +11,27 @@ PipeOpFeatureUnion = R6Class("PipeOpFeatureUnion",
...
 
-      input1 <- inputs[[1]]
-      targets <- input1$data(cols = task$target_names)
+      input1 = inputs[[1]]
+      targets = input1$data(cols = input1$target_names)
 
-      data <- do.call(cbind, c(all_data, list(targets)))
-      db <- as_data_backend(data)
+      data = do.call(cbind, c(all_data, list(targets)))
+      db = as_data_backend(data)
```
pfistfl: Michel proposed new methods in (https://github.com/mlr-org/mlr3/issues/61)
maybe we can use them here?
```
task$select(task$target_names)
task$cbind(newdata)
```

R/Graph.R
```diff
@@ -206,3 +214,66 @@ graph_plot = function(root) {
...
+
+  visit = function(node, state, depth = 1) {
+    if(node$id %in% state$permanent) {
+      state$depth[node$id] = pmax(depth, state$depth[node$id], na.rm = TRUE)
+      return()
+    }
+    if(node$id %in% state$temporary) stop("Not a DAG")
+    state$temporary = c(node$id, state$temporary) # mark temporarily
+
+    if(length(node$next_nodes) > 0) {
```
pfistfl: Do we want a helper function here, so we consistently check this across multiple functions that require this.

zzawadz: I added an AB method `has_next`.

R/PipeOpFeatureUnion.R
zzawadz: I think we can change that when we will have the whole flow of tasks described.

Right now creating new task is safer, because if the same task will be passed to two nodes we won't break anything.

And changing the behavior here was not the purpose of this pull request. I just replaced the `<-` with `=`:)

Issue #37, opened 2018-11-16
Test & verify that %>>% does a deep copy


berndbischl:

zzawadz:
In my implementation in https://github.com/mlr-org/mlr3pipelines/pull/50, it does not copy the inputs.

I have a mixed feeling about this, because if it does a deep copy, then it will be easier for users to define multiple pipelines based on the same building block. So this approach, I think, would be more experimenting friendly.

However, if we wouldn't copy the inputs, then it will be much easier to debug the pipeline because the user will be able to see the current state of the pipeop without looking for it inside the Graph.

Debugging without the deep copy:
```r
g = op1 %>>% op2 %>>% op3
op2$params
```

With deep copy:
```r
g = op1 %>>% op2 %>>% op3
g$find_by_id(op2$id)$pipeop$params
```

On the other hand, I think R users do not work a lot with reference classes, so performing a deep copy will be much more friendly for them.

**I would vote for the deep copy.**

berndbischl:
we need to unit test that it really does a deep copy then close here

PR #50, opened 2018-11-21
Dev pipeoperator

zzawadz:
First attempt to the sugar operators.


zzawadz:
See #51

zzawadz:
#37 - also important.

zzawadz:
I'll add #54 today.

zzawadz:
Now working on #51

zzawadz:
We can merge this if you like the code, but I think a lot of that code might be invalid after we agree on #57

PR #114, opened 2019-01-19
Vignettes and README

mb706:


mb706:
Going to fix trafis in master later today.

PR #137, opened 2019-02-17
Use tic for CI

pat-s:
- use `tic.R`
- adjust to `mlr-org` standard for CI
- use README.md

Appveyor fails due to `igraph` -> as long as there is no binary for the latest release, it will be hard to install on Windows: https://github.com/igraph/rigraph/issues/321
However, once the binary exists the build should pass -> I recommend to include it.


mllg:
Any ideas why roxygen is not getting installed?

pat-s:
The preparation of the `step_build_pkgdown` is missing. In this step, `roxygen2` gets installed. And the reason for missing it is because the env variable `id_rsa` is missing in the Travis settings of the repo.

The reason `travis::use_tic()` and `travis::use_travis_deploy()` did not work (their goal is, among others, to add an `id_rsa` env variable) is that Travis changed from Travis.org to Travis.com for new repos recently. `mlr3` is still in Travis.org while mlr3pipelines is running under Travis.com.

pat-s:
See this issue: https://github.com/ropenscilabs/travis/issues/77

pat-s:
@mllg Merge?

PR #167, opened 2019-06-27
Plotting using Add visNetwork

pfistfl:
Added interactive (d3.js) plots using visNetwork. 

This is by no means finished, but to stimulate discussion. 

Some plots: 
visNetwork

stacking

bagging


pat-s:
Thanks @pfistfl. Could you also include a short reprex how one of these graphs was generated?

I think the coloring is suboptimal, there should be colors distinguishing different parts by default. Also blue outlines on blue boxes do not look good. But I guess that's all configurable?

pfistfl:
Generally, this would just be the `graph$plot(TRUE)` function for the three graphs defined in the `examples` vignette. 

In general everything is configurable.
Takeaways: 
- Input and Output have a (fixed?) color? Proposals? 
- Outline: Different color from fill, any proposals?

pfistfl:
Example 1:
``` r
  library(mlr3)
  library(mlr3pipelines)
  task = mlr_tasks$get("iris")
  train.idx = sample(seq_len(task$nrow), 120)
  test.idx = setdiff(seq_len(task$nrow), train.idx)
  single_pred = PipeOpSubsample$new() %>>%
    PipeOpLearner$new(mlr_learners$get("classif.rpart"))
  pred_set = greplicate(single_pred, 10L)
  pred_set$plot()
```

<sup>Created on 2019-06-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

pat-s:
With your branch your example looks like this for me:

image

> Input and Output have a (fixed?) color? Proposals?

dark-green?

> Outline: Different color from fill, any proposals?

black outline, white fill?

pfistfl:
If you have the latest commits, 
the $plot() method should render to html per default.
Otherwise use `.$plot(TRUE)`

pfistfl:
Currently it looks like this:
image

mb706:
Maybe make "input" and "output" less prominent (i.e. don't make them bigger than pipeop nodes, and put them in a faded colour), they are pseudonodes after all. You could also experiment with making nodes a different colour depending on their input / output, so that a preprocessing pipeop (task -> task; task -> task) looks different from a learner (task -> null; task -> prediction) or an ensemble (prediction -> prediction). The most common of these combinations could be hardcoded, others could get colors depending on a hash value or something. (`get_class_hierarchy()` could help here). Might be interesting to do the same to the arrows, i.e. colour them according to the data type that flows along them.

pfistfl:
@mb706 
Good idea!
I would suggest doing this in a second iteration?

pat-s:
I like the [graphs](https://ropenscilabs.github.io/drake-manual/plans.html) in the _drake_ package. Maybe that can give some inspiration - don't know which package is doing the graph work there.

pfistfl:
They seem to use both [networkD3](https://www.htmlwidgets.org/showcase_networkD3.html) and `visNetwork` (the one I am using) 

The plot looks like `visNetwork` I guess

pfistfl:
- Added alpha for the colors, 
- Info on hover is now basically `print(pipeop)`
- html is off per default? 

I think we might want to postpone further improvements until after release.

PR #172, opened 2019-07-03
use mlr3misc register_dictionary

mb706:
Uses mlr-org/mlr3misc#21


mllg:
We agreed on just registering stuff directly in the dict.

PR #175, opened 2019-07-08
Allow for types other than classif refr

pfistfl:


pfistfl:
Deactivated checks here, they were not working with mlr3ordinal.
Currently do not see why we would need them.

PR #186, opened 2019-07-14
Less `innum` / `outnum` in PipeOps: varargs (input) and automatic copy (output)

mb706:
1) Instead of having to use the `PipeOpCopy` in some places, it is now possible to add multiple edges to a single output channel; outputs are then copied automatically.
    ```r
    "scale" %>>% PipeOpCopy$new(outnum = 2) %>>% gunion(list("pca", "subsample"))
    ```
    can then be written as
    ```r
    "scale" %>>% gunion(list("pca", "subsample"))
    ```
2) Instead of having to initialize some `PipeOp`s with `innum`, one could now declare one `input` channel as `vararg` by naming it `"..."`. Then multiple edges can be connected to it.
    ```r
    gunion(list("scale", "pca")) %>>% PipeOpFeatureUnion$new(innum = 2)
    ```
    can then be written as
    ```r
    gunion(list("scale", "pca")) %>>% "featureunion"
    ```
    (for this the `PipeOpFeatureUnion` needs to be adjusted to have a `"..."` input channel)

### DIFF
--- a/R/Graph.R
+++ b/R/Graph.R
@@ -199,12 +199,11 @@ Graph = R6Class("Graph",
           src_id, types_src$predict, dst_id, types_dst$predict)
       }
 
-      bad_rows = (self$edges$src_id == src_id & self$edges$src_channel == src_channel) |
-        (self$edges$dst_id == dst_id & self$edges$dst_channel == dst_channel)
+      bad_rows = (self$edges$dst_id == dst_id & self$edges$dst_channel == dst_channel & self$edges$dst_channel != "...")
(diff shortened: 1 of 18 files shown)


pfistfl:
I really like those changes.
Still Not convinced about using char Strings though

berndbischl:
> ```r
> gunion(list("scale", "pca")) %>>% "featureunion"
> ```

how about shortening even the first one to
gunion("scale", "pca")
or
list("scale", "pca")

?

PR #233, opened 2019-08-12
Style pkg

pat-s:
Applying `styler::mlr_style()`


pat-s:
@mb706 	Can we merge this? Build errors are the same as in master.

mb706:
Too many conflicts to merge; instead I adjusted things manually now. I left things I disagree with out (short `if ()` on one line, `on.error({})` on one line, two spaces before inline comment, cosmetic empty lines).

pat-s:
The we should probably discuss the style guide again, otherwise we do not need one.

berndbischl:
> The we should probably discuss the style guide again, otherwise we do not need one.

we will dicuss this in one of the next core meeting but @pat-s @mb706 @mllg must all be present

PR #238, opened 2019-08-19
Created new branch for PipeOpSmote

susanne-207:
Added new R file for PipeOpSmote operator

### DIFF
--- a/R/PipeOpSmote.R
+++ b/R/PipeOpSmote.R
@@ -1,0 +1,114 @@
+#' @title PipeOpSmote
+#'
+#' @usage NULL
+#' @name mlr_pipeops_smote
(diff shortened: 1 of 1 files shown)


pfistfl:
Have we decided against using Bernd's implementation? 
See [mlr c implementation](https://github.com/mlr-org/mlr/blob/master/src/smote.c)

mb706:
We can not depend on mlr, and I'd prefer to use other packages before we maintain stuff ourselves, even if that stuff is copied from another library.

pfistfl:
I see, not sure whether just copying the .c file and like 100 lines of code over from mlr might not be the better option. There we at least know that the implementation is solid and well-tested. (and efficient, as it is written in c)

PR #269, opened 2019-09-25
fix_factors_prediction PipeOp

mb706:


mb706:
apparently this is not how it should work, so using the other method instead.

Issue #262, opened 2019-09-17
Learner based imputation PipeOp


mb706:
Should be easy with the neat `PipeOpImpute` base class.

prockenschaub:
Would this simply be a single imputation based on mean predictions by a learner? With a little guidance to keep in line with the overall package design I would be happy to take that on.

Could this be extended to multiple imputation? I need to include MI in my modelling workflow and would be very interested to see/create a PipeOp for this. This would most likely result in a branched workflow with one flow per imputed dataset, which later will have to be "unbranched"?

mb706:
Yes, the plan is to inherit from [PipeOpImpute](https://github.com/mlr-org/mlr3pipelines/blob/master/R/PipeOpImpute.R) with a `PipeOp` that is constructed with a `Learner` and to some degree works similarly to `PipeOpLearner`. I think this would be quite an advanced thing to implement properly; if you feel up to the task you can try it, otherwise I would propably get around to it within maybe a month.

MI is an interesting Idea; if the `Learner` is inherently stochastic it would be possible to do something
```r
po("copy", n) %>>%
  greplicate(
    po("imputelrn", imputelearner) %>>%
    po("learner", predictlearner),
  n) %>>%
  po("classif.avg")
```
It may be useful to introduce noise according to `"prob"` or `"se"` prediction, that could be a `PipeOp` as well.

prockenschaub:
I gave it a first shot and created a `PipeOpImputeLearner` class. 

My implementation is modelled very similar to `PipeOpImputeMean` but uses `train` and `predict` of a `Learner` to obtain the mean instead of `mean(feature, na.rm = TRUE)`. The learner acts on the context (all variables indicated as predictors for the imputation), which we use to create a temporary imputation task.

```r
train_imputer = function(feature, type, context) {
    context$impute_col = feature
    task = TaskRegr$new(id = "taskimpute", backend = context, target = "impute_col")
    
    private$.learner$train(task)
    pred = private$.learner$predict(task)
    pred   # Return a prediction object instead of a single mean
}
```

The learner is passed as second in the initializer function and stored as a private variable, similar to what happens in `PipeOpLearner`. The initializer of `PipeOpImp` is called with `whole_task_dependent = TRUE` to provide the context in `train_imputer`.

```r
initialize = function(id = "imputelearner", learner, param_vals = list()) {
    private$.learner = as_learner(learner)$clone(deep = TRUE) 
[...]

PR #290, opened 2019-10-31
First implementation of PipeOpImputeLearner

prockenschaub:
**Reference Issues/PRs**
Create a PipeOp for learner based imputation (#262)

**What does this implement/fix?** 
This is a first try to create an imputation `PipeOp` that uses mlr3 learners to do imputation. The class is called `PipeOpImputeLearner` and inherits from `PipeOpImpute`. 

The internal structure is inspired by `PipeOpImputeMean` and `PipeOpLearner` and can be seen as a hybrid of the two. Simlar to `PipeOpLearner`, `PipeOpImputeLearner` takes a learner as initialisation parameter. Based on the task type supported by that learner (regression or classification),  `PipeOpImputeLearner` then allows to either impute numerical or categorical features. 

As of yet, the implementation does not allow for any iteration, i.e. initialise missing values with values from the marginal distribution and then iteratively impute one column with (imputed and non-missing) values of all the other columns. Instead, either a learner that can handle missing data needs to be used (e.g. `lrn("regr.rpart"`) or only non-missing columns have to be specified as context columns (see example below).

Example:
```R
library("mlr3")
library("mlr3pipelines")

task = tsk("pima")
task$missings()
# diabetes      age  glucose  insulin     mass pedigree pregnant pressure  triceps 
#       0        0        5      374       11        0        0       35      227

# Learner that can handle missing values
po = po("imputelearner", learner = lrn("regr.rpart"))
new_task = po$train(list(task = task))[[1]]
new_task$missings()
# diabetes      age pedigree pregnant  glucose  insulin     mass pressure  triceps 
#       0        0        0        0        0        0        0        0        0 

# Learner that can't handle missing values
non_miss = task$feature_names[task$missings()[-1] == 0]
po = po("imputelearner", learner = lrn("regr.lm"), 
              param_vals = list(context_columns = selector_name(non_miss)))
new_task = po$train(list(task = task))[[1]]
new_task$missings()
# diabetes      age pedigree pregnant  glucose  insulin     mass pressure  triceps 
#       0        0        0        0        0        0        0        0        0 
```

**Other comments?**
The current implementation still has the following issues, any input welcome:
* I haven't found an easy way to impute features with exclusively missing values (compared to `PipeOpImputeMean`, where those features are imputed with `0`), as this would require a learner that always returns a constant (`0` for numbers, `missing`, `mode` or `reference category` for categorical values). Other options would be to return a `NULL` model during training and deal with this in an if-else manner in the `impute` function. Do we need to do this?
* I tried to include tests into test_pipeop_impute.R. I omitted those here, because I wasn't able to solve an issue with deep clone. The default deep clone does not seem to deep clone the `PipeOp`'s state but that is where the fitted learners are located. Any advise on how to deal with this?

Finally, some of the design choices I have made might be suboptimal. I am happy for any recommendations on how to improve those.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -102,6 +102,7 @@ Collate:
     'PipeOpICA.R'
     'PipeOpImpute.R'
     'PipeOpImputeHist.R'
+    'PipeOpImputeLearner.R'
     'PipeOpImputeMean.R'
     'PipeOpImputeMedian.R'
     'PipeOpImputeNewlvl.R'
--- a/R/PipeOpImputeLearner.R
+++ b/R/PipeOpImputeLearner.R
@@ -58,7 +58,6 @@ PipeOpImputeLearner = R6Class("PipeOpImputeLearner",
                            public = list(
                              initialize = function(id = "imputelearner", learner, param_vals = list()) {
                                private$.learner = as_learner(learner)$clone(deep = TRUE)  # FIXME: use `clone=TRUE` when mlr-org/mlr3#344 is fixed
-                               private$.imp_tsk = if(private$.learner$task_type == "regr") TaskRegr else TaskClassif
 
                                super$initialize(id, param_vals = param_vals, whole_task_dependent = TRUE)
                              },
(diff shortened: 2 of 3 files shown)


berndbischl:
I would like to discuss this pr with someone before we merge it. I didn't really like what we did in mlr2. And thanks a lot this seems like a pr with high effort and good communication.

pfistfl:
@prockenschaub 
We have talked to @berndbischl about your PR. I'd like to finish up this PR with his input, is that ok for you? 
Sorry for taking quite a while, hopefully, we can finish this up soon :)

Open Problem:
- For the deep clone issue, we will use the approach used in `PipeOpLearner`
- For the full NA problem, we might want to find a globally well defined solution, most likely imputing `0`
- We should allow all learners here, but need to check that the learner can handle NA's

prockenschaub:
That's perfectly fine with me, no worries!

pfistfl:
Closing here, continued in #386

Issue #303, opened 2020-01-20
ExtractDateFeatures PipeOp


pfistfl:
We currently have no real way of dealing with columns of type `POSIXct`.
A set of features (+ implementation) that can be extracted from a `POSIXct` column can be found in 
[fast.ai](https://github.com/fastai/fastai/blob/master/fastai/tabular/transform.py).

**TODO: Implement a PipeOpDateFeatures**

Functionality:
Extracts a set of `numeric / logical / factor` features from a `POSIXct` column.
- Features to extract:
  - See fast.ai link
  - hour / minute / second
  - night/day, i.e. whether time (or hour) is in [6:00 am - 8:00pm] or not
  - other ideas?

Params: 
- A column name to operate on. (Check whether this is of the correct type).
 ~~A logical flag for each date feature (whether this should be extracted or not)~~ # maybe do this later
- ~~A logical flag: Whether to remove constant features (default: TRUE)~~

Caveats:
- Try to not introduce additional (or as little as possible) package dependencies.
- Removing constant features: Extract all functions, check whether any columns are constant.
  Make sure, that the colnames extracted are equal in `train/predict`, i.e. we have to store a 
  vector of `colnames` used in `train` to only extract those features during prediction.
  `PipeOpRemoveConstants` can guide you there.

Open Question:
Do we want to have removing constants embedded in this `PipeOp` or just force the user to always
use `PipeOpRemoveConstants` afterwards.

sumny:
Should this only work for objects of class `Date`? If we directly implement this for objects of class `POSIXct`, `Dates` are supported but also hours, minutes and seconds would be available as features.

pfistfl:
Currently `mlr3` supports dates of type `pxc = "POSIXct"`. 
I would assume we should work on it. 
I will update the specs above.

sumny:
So I drafted an early version that can be found in the `pipeop_datefeatures` branch.
Currently supports:
  - year
  - month
  - week of the year
  - day of the year
  - day of the month
  - day of the week
  - hour
  - minute
  - second
  - night/day (whether hour is in 6.00 am - 08.00pm)

Currently, all engineered features are numeric except for `is_day` which is logical. I guess this needs some discussion, i.e., most of them could also be ordered factors? I think fast.ai simply returns numeric features (?)

[...]

PR #308, opened 2020-01-30
Pipeop datefeatures

pfistfl:


R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+        ParamLgl$new("week_of_year", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("day_of_year", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("day_of_month", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("day_of_week", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("hour", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("minute", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("second", default = TRUE, tags = "date_feature"),
+        ParamLgl$new("is_day", default = TRUE, tags = "date_feature")
+      ))
+      ps$values$date_variable = "date" # needed for mlr_pipeops overview
```
pfistfl: Do we have a convention here? Should this be NULL or something?

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+#' @family PipeOps
+#' @include PipeOpTaskPreproc.R
+#' @export
+PipeOpDateFeatures = R6Class("PipeOpDateFeatures",
+  inherit = PipeOpTaskPreprocSimple,
+  public = list(
+    initialize = function(id = "datefeatures", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("date_variable", custom_check = check_character, tags = "required"),
+        ParamLgl$new("cyclic", default = FALSE),
```
pfistfl: tags should be `train`, `predict`, ... not `date_feature`

R/PipeOpDateFeatures.R
pfistfl: Also I would just initialize all logical Params above to `TRUE`here.
i.e.
`ps$values = list(xyz = TRUE, xyz2 = TRUE, ...)`

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+#' pop$state
+#' @family PipeOps
+#' @include PipeOpTaskPreproc.R
+#' @export
+PipeOpDateFeatures = R6Class("PipeOpDateFeatures",
+  inherit = PipeOpTaskPreprocSimple,
+  public = list(
+    initialize = function(id = "datefeatures", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("date_variable", custom_check = check_character, tags = "required"),
```
pfistfl: I guess we also want a `keep_date_variable` that lets us decide whether we want to drop the `date_variable` after computing the features. Defaults to `FALSE`

pfistfl: Additionally, I think I would rename `date_variable` to either `date_var` or `date_col`.

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
+#' @title PipeOpDateFeatures
+#'
+#' @usage NULL
+#' @name mlr_pipeops_datefeaturs
+#' @format [`R6Class`] object inheriting from [`PipeOpTaskPreprocSimple`]/[`PipeOp`].
+#'
+#' @description
+#' Based on a specified `POSIXct` column of the data, a set of date related features is computed and
+#' added to the feature set of the output task.
+#'
```
pfistfl: Please add a reference to the original source code from fast.ai

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+#' The output is the input [`Task`][mlr3::Task] with date-related features computed and added to the
+#' feature set of the output task and the specified `POSIXct` column of the data removed from the
+#' feature set.
+#'
+#' @section State:
+#' The `$state` is a named `list` with the `$state` elements inherited from
+#' [`PipeOpTaskPreprocSimple`], as well as:
+#' * `date_variable` the name of the `POSIXct` column that was used for the feature engineering.
+#' * `cyclic` logical value indicating whether cyclic features were computed.
```
pfistfl: ... should be computed ...

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+#' * `features` a vector of the feature names that were computed.
```
pfistfl: vector of computed feature names

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+#' * `minute` :: `logical(1)`\cr
+#'   Whether the minute should be extracted as a feature. Default TRUE.
+#' * `second` :: `logical(1)`\cr
+#'   Whether the second should be extracted as a feature. Default TRUE.
+#' * `is_day` :: `logical(1)`\cr
+#'   Whether a feature should be extracted indicating whether it is day time (06:00 am - 08:00pm).
+#'   Default TRUE.
+#'
+#' @section Internals:
+#' If `cyclic = TRUE`, cyclic features are computed for the features `"month"`, `"week_of_year"`,
```
pfistfl: are computed ... ? in addition to ? ...

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,209 @@
...
+      cyclic = ifelse(is.null(self$param_set$values$cyclic), yes = self$param_set$default$cyclic,
+        no = self$param_set$values$cyclic)
+      list(date_variable = date_variable, cyclic = cyclic, features = features)
+    },
+
+    transform = function(task) {
+      dates = task$data()[[self$state$date_variable]]
+      assert_posixct(dates)
+      features = self$state$features
+      date_features = as.data.table(lapply(features, function(feature) {
```
pfistfl: Can you use more idiomatic `data.table` syntax to make things mroe legible?
https://stackoverflow.com/questions/11680579/assign-multiple-columns-using-in-data-table-by-group

pfistfl:
Thanks a lot for the PR! I made a few comments, but in general, I like it a lot!

Currently missing are unit tests, other than that I am okay!

R/PipeOpDateFeatures.R
pfistfl: So currently, we overwrite the date features when we compute cyclic features?
- Is this how it is done in fast.ai?
- Do we want to keep this?

R/PipeOpDateFeatures.R
sumny: in the updated version I initialize to `character(0L)` but I guess this is still not optimal?

R/PipeOpDateFeatures.R
sumny: fixed

R/PipeOpDateFeatures.R
sumny: fixed

R/PipeOpDateFeatures.R
sumny: fixed

R/PipeOpDateFeatures.R
sumny: fixed

R/PipeOpDateFeatures.R
sumny: fixed

R/PipeOpDateFeatures.R
sumny: I changed this in the updated version so that cyclic features are kept as additional ones.

R/PipeOpDateFeatures.R
sumny: fixed (?)

sumny:
Thanks for the feedback! I agree with most of your comments and made appropriate changes. Also, I added unit tests. Regarding how to initialize `date_var` I am still unsure. Currently this defaults to `character(0L)` but then I cannot rely on `expect_datapreproc_pipeop_class` in the unit tests.

tests/testthat/test_pipeop_datefeatures.R
```diff
@@ -0,0 +1,88 @@
...
+  expect_identical(train_pipeop(po, inputs = list(task))$output$feature_names,
+    c("Petal.Length", "Petal.Width", "Sepal.Length", "Sepal.Width",
+      "month", "week_of_year", "day_of_year", "day_of_month", "day_of_week",
+      "hour", "minute", "is_day",
+      "month_sin", "month_cos", "week_of_year_sin", "week_of_year_cos",
+      "day_of_year_sin", "day_of_year_cos", "day_of_month_sin", "day_of_month_cos",
+      "day_of_week_sin", "day_of_week_cos",
+      "hour_sin", "hour_cos", "minute_sin", "minute_cos"))
+})
+
```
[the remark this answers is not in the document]
pfistfl: - Check NA for cyclic
- Check that something reasonable happens for constant dates

tests/testthat/test_pipeop_datefeatures.R
```diff
@@ -0,0 +1,127 @@
+context("PipeOpDateFeatures")
+
+dat = iris
```
pfistfl: remove those lines

pfistfl:
Looks good!
@sumny is done here, I will take over merging!

sumny:
I found a bug and fixed it: For cyclic features, maxima must be with respect to values starting at 1, not 0:
`hours = 0:23`
`hours_sin = sin(2 * pi * hours / 24)`
`hours_cos = cos(2 * pi * hours / 24)`
`cols = qualitative_hcl(24)`
`par(mfrow = c(1, 2))`
`plot(hours, type = "n")`
`text(hours, col = cols, as.character(hours))`
`plot(hours_sin, hours_cos, type = "n")`
`text(hours_sin, hours_cos, col = cols, as.character(hours))`
Everything else should be fine.

mb706:
Hey, I looked into this a bit and made a few corrections.

There also seems to be a small bug concerning the cyclic features? When I set every parameter to FALSE except year and have 'cyclic = TRUE` I get a warning:
```r
> dat = iris
> set.seed(1)
> dat$date = sample(seq(as.POSIXct("2020-01-31"), to = as.POSIXct("2020-03-01"), by = "sec"), 150L)
> task = TaskClassif$new("iris_date", backend = dat, target = "Species")
>    po = PipeOpDateFeatures$new(param_vals = list(date_var = "date"))
> po$param_set$values$is_day = FALSE
> po$param_set$values$second = FALSE
> po$param_set$values$minute = FALSE
> po$param_set$values$hour = FALSE
> po$param_set$values$day_of_week = FALSE
> po$param_set$values$day_of_month = FALSE
> po$param_set$values$day_of_year = FALSE
> po$param_set$values$week_of_year = FALSE
> po$param_set$values$month = FALSE
> po$param_set$values$cyclic = TRUE
> po$train(list(task))
$output
<TaskClassif:iris_date> (150 x 6)
* Target: Species
* Properties: multiclass
* Features (5):
  - dbl (5): Petal.Length, Petal.Width, Sepal.Length, Sepal.Width, year

Warning messages:
1: In `[.data.table`(dates, , `:=`(paste0(rep(cyclic_features, each = 2L),  :
  Column '_sin' does not exist to remove
2: In `[.data.table`(dates, , `:=`(paste0(rep(cyclic_features, each = 2L),  :
  Column '_cos' does not exist to remove
```

mb706:
I also think that this PipeOp should be able to work on multiple date features at once. I suggest following changes:
* remove the `date_var` parameter, instead operate on all `POSIXct` or compatible date features. If the user wants to operate only on a particular date feature, he can use the the `affect_columns` functionality
* call the `compute_date_features` function multiple times for each date feature being operated on, creating columns of the form `"[orig feat name].year"`, `"[orig feat name].week"` etc.
* Remove the warning about not modifying any data.
* I think overloading the `get_state` function is superfluous. `PipeOpTaskPreproc` does the heavy lifting of ensuring the same features (possibly in different order, but this should not be a problem) are present during training and prediction. You can also rely on `param_set$values` not changing between training and prediction.
* Therefore add the "predict" tag to all parameters.
* I think it is a good idea to use `transform_dt` instead of `transform` and operate only on the `data.table` of features. Instead of having to sniff out the `POSIXct` features you can then overload the `select_cols` function. Use e.g. PipeOpSpatialSign as an example (which selects integers and numerics instead of POSIXct but the transfer is obvious).
* Don't do the `setdiff(self$param_set$ids(...), c(..., "cyclic")` thing to get a bunch of parameters. You always want to get parameters by "white list", not "black list", because PipeOpTaskPreproc may add a `Lgl` parameter in the future and then your code breaks. Instead give all the part-of-date-to-extract parameters a custom tag and get parameters by that tag (`self$param_set$ids(tags = "datepart")`). See e.g. PipeOpYeoJohnson how to do this.

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,240 @@
+#' @title PipeOpDateFeatures
+#'
+#' @usage NULL
+#' @name mlr_pipeops_datefeaturs
```
mb706: Typo!

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,240 @@
...
+#' @format [`R6Class`] object inheriting from [`PipeOpTaskPreprocSimple`]/[`PipeOp`].
```
mb706: Should be ``[`PipeOpTaskPreprocSimple`]/[`PipeOpTaskPreproc`]/[`PipeOp`]``

pfistfl:
```
Instead give all the part-of-date-to-extract parameters a custom tag and get parameters by that tag (self$param_set$ids(tags = "datepart")). See e.g. PipeOpYeoJohnson how to do this.
```

We do this now? Sorry, I did not know that.

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,226 @@
...
+      return((6L <= hours) & (hours <= 20L)) # early exit
+    }
+    as.numeric(format(dates, formatting))
+  })
+}
+
+# helper function to compute cyclic date features of date features, i.e.,
+# sine and cosine transformations of 2 * pi * x / max_x + 1
+compute_cyclic_date_features = function(date_features, features, date_var) {
+  do.call(c, args = lapply(features, FUN = function(feature) {
```
mb706: 1. *always* use `base::c` and not `c` when referring to the `c` function but don't have parentheses. (`c()` always looks up the `c`-function, but just `c` could also be a variable named `c`, and that name is so short that this easily hapens.
2. don't do `do.call(c`, do `unlist(..., recursive = FALSE)`.

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,226 @@
...
+#'   Should the second be extracted as a feature? Default TRUE.
+#' * `is_day` :: `logical(1)`\cr
+#'   Should a feature be extracted indicating whether it is day time (06:00am - 08:00pm)?
+#'   Default TRUE.
+#'
+#' @section Internals:
+#' If `cyclic = TRUE`, cyclic features are computed for the features `"month"`, `"week_of_year"`,
+#' `"day_of_year"`, `"day_of_month"`, `"day_of_week"`, `"hour"`, `"minute"` and `"second"`. This
+#' means that for each feature, `x`, two additional features are computed, namely the sine and
+#' cosine transformation of `2 * pi * x / max_x + 1`. This is useful to respect the cyclical nature
```
mb706: I think the documentation should be clearer here:

1. what is `max_x`?
2. Why the `+ 1`?

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,226 @@
...
+      hours = as.numeric(format(dates, formatting))
+      return((6L <= hours) & (hours <= 20L)) # early exit
+    }
+    as.numeric(format(dates, formatting))
+  })
+}
+
+# helper function to compute cyclic date features of date features, i.e.,
+# sine and cosine transformations of 2 * pi * x / max_x + 1
+compute_cyclic_date_features = function(date_features, features, date_var) {
```
mb706: is there a more elegant way to get to the date values instead of using `date_features[[paste0(date_var, ".", XXX)]]` all the time? Maybe just give the `date_var.feature` column as an argument directly?

R/PipeOpDateFeatures.R
```diff
@@ -0,0 +1,241 @@
...
+  (year %% 4L == 0L & year %% 100L != 0L) | (year %% 400L == 0L)
+}
+
+# helper function to get the number of days per month respecting leap years
+get_days_per_month = function(year, month) {
+  stopifnot(length(year) == length(month))
+  if (length(year) > 1L) {
+    mapply(get_days_per_month, year = year, month = month)
+  } else {
+    days_per_month = c(31L, 28L, 31L, 30L, 31L, 30L, 31L, 31L, 30L, 31L, 30L, 31L)
```
mb706:
```r
days_per_month = c(31L, 28L, 31L, 30L, 31L, 30L, 31L, 31L, 30L, 31L, 30L, 31L)
if (is_leap_year(year)) days_per_month[2] = 29
days_per_month[month]
```
does the NA handling automatically

sumny: if year is `NA` like in the test"PipeOpDateFeatures - automatic NA handling" this does not work because is_leap_year(year) will return `NA` but I get the point - fixed in the next commit

mb706: Hm, why can 'year' and 'month' be NA at all? Also what *should* the year == NA behaviour be? february with 28 days? 28.25? Maybe just change the value of is_leap_year(NA) to FALSE?

sumny: `year` and `month` can be NA if the according `POSIXct` value is a complete `NA` (this can happen if a user constructs the `POSIXc` column with `optional = TRUE`) I thought that `PipeOpDateFeatures` should then also return `NA` for the datefeatures, i.e., all functions should simply pass `NA` through. As far as I know, we only have to distinguish between correctly specified `POSIXct` values (then there cannot be any `NA`'s in any fields) or complete `NA` values, then all extracted features must also be `NA`

sumny: so in my opinion there are two possibilities to "fix" this: 1) leave as is, because it handles `NA` correctly 2) use a saver version to check whether a `POSIXct` column is `NA` and if it is, then all features are automatically set to `NA` so we must not further check whether `NA`s get passed through all functions correctly (I guess I prefer the second one)

sumny:
can this be merged? closes #303

pfistfl:
Merge https://github.com/mlr-org/mlr3gallery/pull/13 once this PR hits CRAN

PR #309, opened 2020-01-31
Some fixes to comparison vignette. Still not working

adibender:
- `credit_data` now lives in package `modeldata` instead of `recipes`
- `PipeOpImpute$new()` throws error unless `id` is specified 
- the last call `resample(...)` throws cryptic error 
```
Error in self$train_imputer(col, type, context) : Abstract.
```
???

### DIFF
--- a/vignettes/comparison_mlr3pipelines_mlr_sklearn.Rmd
+++ b/vignettes/comparison_mlr3pipelines_mlr_sklearn.Rmd
@@ -364,10 +364,11 @@ tsk = TaskClassif$new(id = "credit_task", target = "Status",
   backend = as_data_backend(data.table(credit_data)))
 
 # Build up the Pipeline:
-g = PipeOpImpute$new(id = "impute") %>>%
-  # PipeOpEncode$new() %>>%
-  # PipeOpScale$new() %>>%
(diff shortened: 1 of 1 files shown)


pfistfl:
Hey,

AFAIK `PipeOpImpute()` is an abstract base class, this should not be used. 
Use a concrete class, i.e. `PipeOpImputeHist`, `PipeOpImputeConstant`, ... instead.

I hope that fixes your problem, else ping me again.

adibender:
I didn't write the code. I just tried to run it and it didn't work. If it's not supposed to be used, why does it appear multiple times in the vignette?

pfistfl:
I think this was changed after the vignette was created. 
Can we merge this?

adibender:
I will check later today if everything works. In general: Why are all the chunks set to `eval = FALSE`? That way errors will never be caught.

pfistfl:
Because we did not want travis builds to break because of it.
I am also not 100% sure about the quality of this vignette. 
I created this vignette in order to show and compare frameworks, but as I am not an expert in the other frameworks, I am not sure about my ramblings.

pfistfl:
Will not merge this here, merging this into the `mlr3book` asap.

PR #323, opened 2020-02-12
Add pipeline dict

pfistfl:
Introduces a dictionary of often-used pipelines.


R/mlr_graphs.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' @usage NULL
+#' @format [`R6Class`] object inheriting from [`mlr3misc::Dictionary`].
+#'
+#' @description
+#' A simple [`Dictionary`][mlr3misc::Dictionary] storing objects of class [`Graph`].
+#' The dictionary contains a collection of often-used graph structures, and it's aim
+#' is solely to make often-used functions more accessible.
+#' Each `Graph` has an associated help page, which can be accessed via `?<key>`, i.e.
+#' `?bagging_pipeline`.
```
mb706: For the other dictionary elements we usually have `?<dictname>_<key>`, e.g. `mlr_pipeops_colapply`

R/mlr_graphs.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' @section Methods:
+#' Methods inherited from [`Dictionary`][mlr3misc::Dictionary], as well as:
+#' * `add(key, value)`\cr
```
mb706: signatur stimmt nicht

R/mlr_graphs.R
```diff
@@ -0,0 +1,56 @@
...
+#'   (`character(1)`, `R6ClassGenerator`, `NULL` | `list`)\cr
+#'   Adds constructor `value` to the dictionary with key `key`, potentially
+#'   overwriting a previously stored item.
+#'
+#' @section S3 methods:
+#' * `as.data.table(dict)`\cr
+#'   [`Dictionary`][mlr3misc::Dictionary] -> [`data.table::data.table`]\cr
+#'   Returns a `data.table` with column `key` (`character`).
+#' @family mlr3pipelines backend related
+#' @family Graph
```
mb706: does this family have another member? If not I'd probably remove this. I don't think it should be put together with the `?Graph` help page.
Maybe instead make a `@family Dictionaries` and add this and the `mlr_pipeops` dict.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+#'   Maximum number of factor levels allowed. See above. Default: 1000.
+#' @return Returns a [`Graph`]
+#' @export
+#' @examples
+#' library(mlr3)
+#' lrn = lrn("regr.rpart")
+#' task = mlr_tasks$get("boston_housing")
+#' gr = robustify_pipeline(task, lrn) %>>% po("learner", lrn)
+#' resample(task,GraphLearner$new(gr), rsmp("holdout"))
+robustify_pipeline = function(task, learner = NULL, scaling = TRUE, max_cardinality = 1000) {
```
mb706: maybe make task optional and assume worst case if not given?

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+  assert_logical(scaling)
+  assert_count(max_cardinality)
+
+  cols_by_type = function(types) {
+    assert_character(types)
+    task$feature_types[get("type") %in% types,]
+  }
+  pos = list()
+
+  if ("factor" %in% learner$feature_types %??% c() & nrow(cols_by_type("factor")) > 0)
```
mb706: always use `&&` for scalar conditions

mb706: Also the `%??%` part is not necessary here?

R/mlr_graphs_elements.R
mb706: also a suggestion: have the arguments
```r
function(task = NULL, learner = NULL, impute_missings = NULL, factors_to_numeric = NULL, ....)
```
and impute missings if `impute_missings` is `TRUE`, of if (`impute_missings` is `NULL` and (task is `NULL` or has `missings`) and (learner is `NULL` or can not handle missings)); similarly with factor to numeric

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+    # Collapse factors over 1000 levels
+    if (any(map_lgl(task$levels(cols_by_type("factor")$id), function(x) length(x) > max_cardinality)))
+      pos = c(pos, po("collapsefactors", param_vals = list(target_level_count = max_cardinality)))
+    # Encode factors
+    if (nrow(cols_by_type("factor")) > 0)
+      pos = c(pos, po("encode"))
+  }
+
+  if (scaling) pos = c(pos, po("scale"))
+  if (length(pos) == 0) return(po("nop"))
```
mb706: return a Graph with element `nop`, not a pipeop `nop`. this is important in some cases because a pipeop `$train()` behaves differently than a graph `$train()`.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+  else Reduce(`%>>%`, pos)
```
mb706: same here, make sure a graph is returned. we have the awfully named `as_graph()` for that.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+#' @return Returns a [`Graph`]
+#' @export
+#' @examples
+#' library(mlr3)
+#' lrn_po = po("learner", lrn("regr.rpart"))
+#' task = mlr_tasks$get("boston_housing")
+#' gr = bagging_pipeline(lrn_po, 3, averager = po("regravg"))
+#' resample(task, GraphLearner$new(gr), rsmp("holdout"))
+bagging_pipeline = function(graph, iterations = 10, averager = NULL) {
+  assert_true(inherits(averager, "PipeOp") | inherits(averager, "Graph"))
```
mb706: always use `||` in these cases.

mb706: Also use `if (!is.null(averager)) averager = as_graph(averager)`, which checks and converts in one go.

mb706: and if we check that things are graphs, you can just as well also check `graph = as_graph(graph)`

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+  assert_count(iterations)
+
+  subs = po("subsample", param_vals = list(frac = 0.7)) %>>% graph
+  subs_repls = greplicate(subs, iterations)
+
+  if (!is.null(averager)) {
+    assert_true(inherits(averager, "PipeOp") | inherits(averager, "Graph"))
```
mb706: delete this

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+    subs_repls %>>% averager
+  } else {
+    return(subs_repls)
```
mb706: no return

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+#' @param task [`Task`] \cr
+#'   A task to create a robustifying pipeline for.
+#' @param learner [`Learner`] \cr
+#'   A learner to create a robustifying pipeline for. Optional, if omitted,
+#'   a more conservative pipeline is built.
+#' @param scaling [`logical`] \cr
+#'   Should the data be scaled? Defaults to true.
+#' @param max_cardinality [`integer`] \cr
+#'   Maximum number of factor levels allowed. See above. Default: 1000.
+#' @return Returns a [`Graph`]
```
mb706: just ``@return [`Graph`]``

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+#'   a more conservative pipeline is built.
+#' @param iterations [`integer`] \cr
+#'   Number of bagging iterations. Defaults to 10.
+#' @param averager [`PipeOp`]|[`Graph`] \cr
+#'   A [`PipeOp`] or [`Graph`] that averages the predictions from the
+#'   replicated and subsampled graph's.
+#'   In the simplest case, `po("classifavg")` and `po("regravg")` can be used
+#'   in order to perform simple averaging of classification and regression
+#'   predictions respectively.`
+#' @return Returns a [`Graph`]
```
mb706: just ``@return [`Graph`]``

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,118 @@
...
+
+mlr_graphs$add("robustify_pipeline", robustify_pipeline)
+
+
+#' @title Create a bagging learner
+#'
+#' @description
+#' Creates a [`Graph`] that performs bagging on the supplied graph.
+#'
+#' @param graph [`PipeOp`]|[`Graph`] \cr
```
mb706: space before and after `|`

mb706: same at other occurrences

R/pipe.R
```diff
@@ -0,0 +1,28 @@
...
+#' @param .obj `[any]`\cr
+#'   The object from which to construct a `PipeOp`. If this is a
+#'   `character(1)`, it is looked up in the [`mlr_pipeops`] dictionary.
+#'   Otherwise, it is converted to a `PipeOp`.
+#' @param ... `any`\cr
+#'   Additional parameters to give to constructed object.
+#'   This may be an argument of the constructor of the
+#'   `PipeOp`, in which case it is given to this constructor;
+#'   or it may be a parameter value, in which case it is
+#'   given to the `param_vals` argument of the constructor.
```
mb706: missing `@return`

R/pipe.R
```diff
@@ -0,0 +1,28 @@
...
+#' @export
+#' @examples
+#' library("mlr3")
+#'
+#' gr = pipe("bagging_pipeline", graph = po(lrn("regr.rpart")), averager = po("regravg"))
+pipe = function(.obj, ...) {
+  UseMethod("pipe")
```
mb706: I don't think there is a good reason for this; the `po()` accessor has additional functionality of converting filters, learners etc. to pipeops, but there is no natural conversion from things to pipelines, or is there?

pfistfl: removed s3 functionality for now

R/mlr_graphs_elements.R
pfistfl: Why do ` impute_missings = NULL, factors_to_numeric = NULL` default to NULL?

R/mlr_graphs.R
pfistfl: Can do that

pfistfl:
Should `robustify_pipeline` also deal with imbalanced data.

R/mlr_graphs_elements.R
mb706: The idea is that `impute_missings` and `factors_to_numerics` may override the `task` / `learner` properties.

mb706: You could also do something like
```r
function(task = NULL, learner = NULL, 
  impute_missings = (is.null(task) || any(task$missings())) && (is.null(learner) || !"missings" %in% learner$properties),
  factors_to_numeric = (is.null(task) || <TASK_HAS_FACTORS>) && (is.null(learner) || !"factors" %in% learner$properties))
```

mb706:
I think "robustify" should do as little as possible to prevent a Learner from crashing, handling imbalances would actually change the resulting model without need.

pfistfl:
Tests not passing is weird; order of missind and histimpute shuld be reversed.

pfistfl:
We might need to find an alternative for the `pipe` sugar as `base::pipe` already exists.

pfistfl:
This should additionally do the following thins (up for debate):
  - Convert posixct to numeric using new pipeop
  - Auto-convert character to factor? Perhaps improve later with text po's
  - Logical to integer conversion

pfistfl:
I think the tests currently not passing depends on https://github.com/mlr-org/mlr3/issues/437. Otherwise, I would welcome another round of feedback.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+  # FIXME
+  # This should additionally do the following thins:
+  # - Convert posixct to numeric using new pipeop
+  # - Auto-convert character to factor? Perhaps improve later with text po's
+  # - Logical to integer conversion
+
+  # If given a task, only treat actually existing column types
+  if (!is.null(task)) {
+    pos = list()
+    if ("factor" %in% learner$feature_types && nrow(cols_by_type("factor")) > 0)
```
mb706: what is our behaviour if learner can not handle factors? we just ignore new factors because they get converted to numerics? In that case we should go by `factors_to_numeric` here. But maybe we want to fix unseen factors to NAs and then impute in any case? Then just go by `is.null(task) || nrow(cols_by_type("factor"))`

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+      pos = c(pos, po("fixfactors", param_vals = list(affect_columns = is.factor)))
```
mb706: 1. `fixfactors` only affects factor / ordered columns, so `affect_columns` is superfluous here. Unless you want to exclude ordereds for some reason?
2. `is.factor` is a bug, this function ends up selecting 0 columns (and fixing no factors at all). There should probably be a test that would have caught this.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+    if (impute_missings) {
+      # Impute numerics
+      if (nrow(cols_by_type(c("numeric", "integer"))) > 0)
+        pos = c(pos, po("copy", 2) %>>%
```
mb706: we don't need `PipeOpCopy` at this point.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+
+    if (factors_to_numeric) {
+      # Collapse factors over 1000 levels
+      if (any(map_lgl(task$levels(cols_by_type("factor")$id), function(x) length(x) > max_cardinality)))
+        pos = c(pos, po("collapsefactors", param_vals = list(target_level_count = max_cardinality)))
+      # Encode factors
+      pos = c(pos, po("encode"))
+    }
+  # If no task is provided, be conservative
+  } else {
```
mb706: I think everything below here should go away and the `if (!is.null(task))` should also go. Otherwise we have two code paths that do 80% the same thing (and interestingly get bugs in one but not the other code path, see `affect_columns` used wrong in `PipeOpFixFactors` above but not below. Instead the `if`s above should have `is.null(task) || ` when necessary.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+            po("imputehist", param_vals = list(affect_columns = selector_type(c("numeric", "integer")))),
+            po("missind",  param_vals = list(affect_columns = selector_type(c("numeric", "integer")))))) %>>%
+          po("featureunion"),
+        po("imputenewlvl", param_vals = list(affect_columns = selector_type(c("factor", "ordered")))))
+    }
+    pos = c(pos, po("collapsefactors", param_vals = list(target_level_count = max_cardinality)))
+    if (factors_to_numeric) pos = c(pos, po("encode"))
+  }
+
+  if (scaling) pos = c(pos, po("scale"))
```
mb706: is scaling every necessary to make a model robust against crashes? Otherwise this becomes a generic preprocessing pipeline.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+
+  if (length(pos) == 0) as_graph(po("nop"))
+  else Reduce(`%>>%`, pos)
```
mb706: ``Reduce(`%>>%`, pos) `` returns a single `PipeOp` (i.e. not a `Graph`) if `length(pos)` is ever 0. I think a nicer way of doing this is
```r
if (!length(pos)) pos = list(po("nop"))
as_graph(Reduce(`%>>%`, pos)
```

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+      # Impute numerics
+      if (nrow(cols_by_type(c("numeric", "integer"))) > 0)
+        pos = c(pos, po("copy", 2) %>>%
+          gunion(list(
+            po("imputehist", param_vals = list(affect_columns = selector_type(c("numeric", "integer")))),
+            po("missind",  param_vals = list(affect_columns = selector_type(c("numeric", "integer")))))) %>>%
+          po("featureunion"))
+      # Impute factors
+      if (nrow(cols_by_type(c("factor", "ordered"))) > 0)
+        pos = c(pos, po("imputenewlvl", param_vals = list(affect_columns = selector_type(c("factor", "ordered")))))
```
mb706: Imputation with new level is a difficult thing here because it may introduce new levels. A better way to do imputation & imputation would be `po("imputehist") %>>% po("imputenewlvl") %>>% po("fixfactors") %>>% po("imputesample", affect_columns = selector_type(c("factor", "ordered")))`.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+  if (!is.null(task)) {
+    pos = list()
+    if ("factor" %in% learner$feature_types && nrow(cols_by_type("factor")) > 0)
+      pos = c(pos, po("fixfactors", param_vals = list(affect_columns = is.factor)))
+    if (impute_missings) {
+      # Impute numerics
+      if (nrow(cols_by_type(c("numeric", "integer"))) > 0)
+        pos = c(pos, po("copy", 2) %>>%
+          gunion(list(
+            po("imputehist", param_vals = list(affect_columns = selector_type(c("numeric", "integer")))),
```
mb706: don't need the affect_columns here, imputehist only goes on numbers

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+    }
+    pos = c(pos, po("collapsefactors", param_vals = list(target_level_count = max_cardinality)))
+    if (factors_to_numeric) pos = c(pos, po("encode"))
+  }
+
+  if (scaling) pos = c(pos, po("scale"))
+
+  if (length(pos) == 0) as_graph(po("nop"))
+  else Reduce(`%>>%`, pos)
+}
```
mb706: removing constant factors is missing? It should come after `missind`.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,163 @@
...
+    averager = as_graph(averager)
+    assert_true(graph$output$train == averager$input$train | averager$input$train == "NULL")
+    assert_true(graph$output$predict == averager$input$predict)
+  }
+
+  subs = po("subsample", param_vals = list(frac = frac)) %>>% graph
+  subs_repls = greplicate(subs, iterations)
+
+  if (!is.null(averager)) subs_repls %>>% averager
+  else subs_repls
```
mb706: do else with curly braces, or all on one line
```r
 if (!is.null(averager)) subs_repls %>>% averager else subs_repls
```

R/mlr_graphs.R
```diff
@@ -0,0 +1,56 @@
...
+#' gr = mlr_graphs$get("robustify", task = task, learner = lrn) %>>% po(lrn)
+#' # or equivalently
+#' gr = pipe("robustify", task, lrn) %>>% po("learner", lrn)
+#'
+#' # all Graphs currently in the dictionary:
+#' as.data.table(mlr_graphs)
+mlr_graphs = R6Class("DictionaryGraph", inherit = mlr3misc::Dictionary,
+  cloneable = FALSE,
+  public = list(
+    add = function(key, value, info) {
```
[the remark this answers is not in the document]
pfistfl: dropped. artifact from earlier

R/mlr_graphs_elements.R
pfistfl: WIll add a test.

R/mlr_graphs_elements.R
pfistfl: +1

R/mlr_graphs_elements.R
pfistfl: Can you name a case where we need this? 
```
po("imputesample", affect_columns = selector_type(c("factor", "ordered")))
```

R/mlr_graphs_elements.R
pfistfl: I think we might wanna use robustify as something I can plug in front of an `svm` and get sensible results. 
I think I will just have `scaling` default to `FALSE` though.

pfistfl:
Codecov does not recognize things obtained from dict apparently.

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,159 @@
...
+#' task = mlr_tasks$get("boston_housing")
+#' gr = bagging_pipeline(lrn_po, 3, averager = po("regravg"))
+#' resample(task, GraphLearner$new(gr), rsmp("holdout"))
+bagging_pipeline = function(graph, iterations = 10, frac = 0.7, averager = NULL) {
+  assert_count(iterations)
+  assert_number(frac, lower = 0, upper = 1)
+  graph = as_graph(graph)
+  if (!is.null(averager)) {
+    averager = as_graph(averager)
+    assert_true(graph$output$train == averager$input$train | averager$input$train == "NULL")
```
mb706: `||`

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,159 @@
...
+    assert_true(graph$output$predict == averager$input$predict)
```
mb706: `are_types_compatible`

mb706: for all rows

mb706: but `%>>%` checks this in any case

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,159 @@
...
+#' @examples
+#' library(mlr3)
+#' lrn = lrn("regr.rpart")
+#' task = mlr_tasks$get("boston_housing")
+#' gr = robustify_pipeline(task, lrn) %>>% po("learner", lrn)
+#' resample(task, GraphLearner$new(gr), rsmp("holdout"))
+robustify_pipeline = function(task = NULL, learner = NULL, impute_missings = NULL,
+  factors_to_numeric = NULL, scaling = FALSE, max_cardinality = 1000) {
+
+  cols_by_type = function(types) {
```
mb706:
```r
has_type_feats = function(types, if_null = TRUE) {
  if (is.null(task)) if_null else any(types %in% task$feature_types)) or something like this
}
```
this saves a lot of `is.null(task) || (nrow(cols_by_type(c("character"))) > 0`

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,159 @@
...
+    if (is.null(task) || nrow(cols_by_type(c("factor", "ordered", "character"))) > 0)
+      pos = c(pos, po("imputenewlvl"))
+  }
+
+  # Fix extra factor levels
+  if (is.null(task) || nrow(cols_by_type(c("factor", "ordered", "character"))) > 0)
+    pos = c(pos, po("fixfactors"))
+
+  # Ensure all factor levels are encoded during predict.
+  if (impute_missings && (is.null(task) || nrow(cols_by_type(c("factor", "ordered"))) > 0))
```
mb706: also logical; same in line below

R/mlr_graphs_elements.R
```diff
@@ -0,0 +1,159 @@
...
+#' @return [`Graph`]
+#' @export
+#' @examples
+#' library(mlr3)
+#' lrn = lrn("regr.rpart")
+#' task = mlr_tasks$get("boston_housing")
+#' gr = robustify_pipeline(task, lrn) %>>% po("learner", lrn)
+#' resample(task, GraphLearner$new(gr), rsmp("holdout"))
+robustify_pipeline = function(task = NULL, learner = NULL, impute_missings = NULL,
+  factors_to_numeric = NULL, scaling = FALSE, max_cardinality = 1000) {
```
mb706: I think we should remove scaling because (1) most methods that benefit from it have it built in already and (2) not scaling the data does not lead to crashed learners, only to bad models in some cases.

pfistfl:
I guess I got everything @mb706

mb706:
Going to merge when travis is green

PR #349, opened 2020-02-24
first PipeOpProxy draft

### LINKED ISSUE #192: PipeOpProxy
We want to write a PipeOp that basically "encapsulates" another PipeOp or another Graph. It has a parameter `content` that can be set to a PipeOp/Graph; then the PipeOp behaves as if it were this PipeOp/Graph. The PipeOpProxy routes its input through the `content` and returns the output. The nice thing is that `$param_set$values$content` is a parameter and can be changed; this is useful as an alternative to branching.

train/predict should look a bit like
```r
train = function(input) {
content = self$param_set$values$content$clone(deep = TRUE)
if (inherits(content, "Graph")) {
output = content$train(unname(input), single_input = FALSE)
} else {
output = content$train(input)
}
self$state = content
output
}
predict = function(input) {
content = self$state
if (inherits(content, "Graph")) {
output = content$predict(unname(input), single_input = FALSE)
} else {
output = content$predict(input)
}
self$state = content
output
}
```

sumny:
Not sure if the ParamSet of the content `PipeOp` or `Graph` should be included in the ParamSet of this proxy pipeop. Also I will write some more tests. Closes #192


R/PipeOpProxy.R
```diff
@@ -0,0 +1,105 @@
...
+PipeOpProxy = R6Class("PipeOpProxy",
+  inherit = PipeOp,
+  public = list(
+    initialize = function(id = "proxy", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("content", tags = c("train", "predidct", "required"), custom_check = function(x) {
+          cl = class(x)
+          ifelse(test_r6(x, classes = "PipeOp") || test_r6(x, classes = "Graph"), yes = TRUE,
+            no =  sprintf("Must inherit from class 'PipeOp' or 'Graph', but has class%s '%s'",
+              ifelse(length(cl) > 1L, yes = "es", no = ""), paste0(cl, collapse = "','")))
```
mb706: general point: I think we shouldn't waste code on proper pluralization

R/PipeOpProxy.R
```diff
@@ -0,0 +1,105 @@
...
+#' proxy$state$state
+#' @family PipeOps
+#' @include PipeOp.R
+#' @export
+PipeOpProxy = R6Class("PipeOpProxy",
+  inherit = PipeOp,
+  public = list(
+    initialize = function(id = "proxy", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("content", tags = c("train", "predidct", "required"), custom_check = function(x) {
```
mb706: How about `tryCatch({ graph = as_graph(x); if (nrow(graph$output) != 1) "Graph must have only one output" else TRUE }, error = "Must be an object that can be converted to a Graph")`? I'm not sure if this works but would be nice because this also covers things that are auto-converted to `PipeOp`s

R/PipeOpProxy.R
```diff
@@ -0,0 +1,105 @@
...
+          cl = class(x)
+          ifelse(test_r6(x, classes = "PipeOp") || test_r6(x, classes = "Graph"), yes = TRUE,
+            no =  sprintf("Must inherit from class 'PipeOp' or 'Graph', but has class%s '%s'",
+              ifelse(length(cl) > 1L, yes = "es", no = ""), paste0(cl, collapse = "','")))
+        })
+      ))
+      ps$values = list(content = PipeOpNOP$new())
+      super$initialize(id, param_set = ps, param_vals = param_vals,
+        input = data.table(name = "input", train = "*", predict = "*"),
+        output = data.table(name = ifelse(test_r6(ps$values$content, classes = "PipeOp"),
```
mb706: Don't make output dependent on parameter value. Instead give it a parameter similar to `PipeOpCopy`'s that defaults to 1

R/PipeOpProxy.R
mb706: also don't use `ifelse`, it is only for vectorization and a bad habit otherwise. Use `if (length(cl) > 1) "x" else "y"`.

R/PipeOpProxy.R
```diff
@@ -0,0 +1,101 @@
...
+#' @family PipeOps
+#' @include PipeOp.R
+#' @export
+PipeOpProxy = R6Class("PipeOpProxy",
+  inherit = PipeOp,
+  public = list(
+    initialize = function(id = "proxy", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("content", tags = c("train", "predidct", "required"), custom_check = function(x) {
+          tryCatch({ graph = as_graph(x); if (NROW(graph$output) != 1L) "Graph must have only one output" else TRUE },
```
mb706: are there tests that cover this?

R/PipeOpProxy.R
```diff
@@ -0,0 +1,101 @@
...
+        })
+      ))
+      ps$values = list(content = PipeOpNOP$new())
+      super$initialize(id, param_set = ps, param_vals = param_vals,
+        input = data.table(name = "input", train = "*", predict = "*"),
+        output = data.table(name = "output", train = "*", predict = "*")
+      )
+    },
+    train_internal = function(input) {
+      content = self$param_set$values$content$clone(deep = TRUE)
```
mb706: i think its better to do
```r
as_graph(self$param_set$values$content, clone = TRUE)
```

mb706: then we also don't need to do "if inherits..."

R/PipeOpProxy.R
```diff
@@ -0,0 +1,101 @@
...
+        output = content$train(unname(input), single_input = FALSE)
+      } else {
+        output = content$train(input)
+      }
+      self$state = content
+      output
+    },
+    predict_internal = function(input) {
+      content = self$state
+      if (inherits(content, "Graph")) {
```
mb706: also don't need 'if inherits' here if we do as_graph in the function above

R/PipeOpProxy.R
mb706: i.e. tests that get each of the error conditions?

R/PipeOpProxy.R
```diff
@@ -0,0 +1,101 @@
...
+#' @include PipeOp.R
+#' @export
+PipeOpProxy = R6Class("PipeOpProxy",
+  inherit = PipeOp,
+  public = list(
+    initialize = function(id = "proxy", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("content", tags = c("train", "predidct", "required"), custom_check = function(x) {
+          tryCatch({ graph = as_graph(x); if (NROW(graph$output) != 1L) "Graph must have only one output" else TRUE },
+            error = function(error_condition) stop("`content` must be an object that can be converted to a Graph", call. = FALSE))
```
mb706: this should not `stop()`, but only return the string

R/PipeOpProxy.R
sumny: I like this suggestion. However, what about PipeOps that have more than one output, i.e., `as_graph(PipeOpCopy$new(2L))$output`. In general should `content` allow more than one output? (and what about input, because there are some PipeOps with more than one?) The current version now only supports one input and one output as content.

R/PipeOpProxy.R
sumny: I do not think that coercing the `content` to a Graph will always work. Because if `content` is a PipeOp with a vararg input, then coercing to a Graph will cause training and predict to fail, i.e., something like the following will happen:
`gr = as_graph(PipeOpFeatureUnion$new())`
`task = tsk("iris")`
`gr$train(list(task, task), single_input = FALSE)`
So I would say that distinguishing between Graphs and anything else is necessary because of such edge cases. However Learner etc. that _could_ be coerced to Graphs must then be coerced (to Graphs or PipeOps) etc.

sumny:
I tried to work with the possibility of a varag input channel and a flexible output channel etc. This seems to work but always coercing `content` to a Graph is not practiable here (see above). I guess we have to think about the design some more, i.e., handle Graphs, PipeOps and Learners etc. separately? Also the `expect_datapreproc_pipeop_class` test seems to show problems with deep cloning the `content`.. So still WIP

R/PipeOpProxy.R
mb706: This is a bug :-/

mb706:
Can you merge master and see if the vararg stuff works now?

sumny:
> Can you merge master and see if the vararg stuff works now?

Coercing the `content` always to a Graph seems to work fine now (including vararg input channels), thanks! Deep cloning of the `content` is still faulty but I will look into this.

mb706:
I think the solution here should be something like
```r
private = list( ...
deep_clone = function(name, value) {
  if (name == "state") {
    value$content = value$content$clone(deep = TRUE)
    value
  } else {
   super$deep_clone(name, value)
  }
}, ...)
```

sumny:
> I think the solution here should be something like
> 
> ```r
> private = list( ...
> deep_clone = function(name, value) {
>   if (name == "state") {
>     value$content = value$content$clone(deep = TRUE)
>     value
>   } else {
>    super$deep_clone(name, value)
>   }
> }, ...)
> ```

I think this should solve the issue with the state after the PipeOp is trained. Still, there should be an issue with deep cloning the `.param_set` because the `value` of `content` is again an R6 object?

PR #356, opened 2020-02-26
PO Threshold

pfistfl:
PipeOp to set the threshold of a prediction.
Can be tuned from the outside.


pfistfl:
This can be reviewed, ignore the stuff in `attic` for now, this will be required for `PipeOpTuneThreshold` later.

mllg:
This just allows to set the threshold, right? Is there a way to tune the threshold without re-fitting the pipeline?

mllg:
The pipeop should probably have these inputs/outputs?

```r
  input = data.table(name = "input", train = "NULL", predict = "PredictionClassif"),
  output = data.table(name = "output", train = "NULL", predict = "PredictionClassif")
```

pfistfl:
Currently not, as the Leaner does not pass  anything onward during train.
If we change this, we could threshold tune the learner.

PR #358, opened 2020-03-02
Pipeop featureunion check

### LINKED ISSUE #271: PipeOpFeatureUnion should throw an error when the same column is added twice

sumny:
closes #271
However, seems to be in conflict with the design of `PipeOpMissInd` and `PipeOpImpute` if both are used before `PipeOpFeatureUnion` and a `selector_type` is specified. Then both
PipeOps can send the same columns through and FeatureUnion cannot unite, see
`test_mlr_graphs.R:73: error: Robustify Pipeline` in the PR


pfistfl:
See #359, this should fix stuff for `PipeOpMissInd`, as only new features with suffix `_missing` should make it through. This should fix your problem I think.
Wanna merge #359 into this PR and see?

sumny:
> See #359, this should fix stuff for `PipeOpMissInd`, as only new features with suffix `_missing` should make it through. This should fix your problem I think.
> Wanna merge #359 into this PR and see?

Thanks for the hint! After merging PR 359 less features are duplicates, however, it seems that factors still get passed through? `test_mlr_graphs.R:73: error: Robustify Pipeline`; `chas` and `town` are still duplicates

pfistfl:
The culprit here seems to be L 177-181 + L192 from `PipeOpTaskPreproc.R`

Problem: Not-selected cols (fct,chr, ...) are added back in after transformation. 

I guess we either need to rewrite `PipeOpMissInd`, so it does not call `train_internal` anymore.
The other big question is how do we expect this to behave, i.e. do we want to guarantee that inheriting from TaskPreproc always means that all non-affected columns are passed on?

pfistfl:
See issue:
- If name clashes occur, compare columns. If values agree, keep one, else error

R/PipeOpMissingIndicators.R
```diff
@@ -99,7 +99,7 @@ PipeOpMissInd = R6Class("PipeOpMissInd",
     .transform = function(task) {
       if (!length(self$state$indicand_cols)) {
         # need to handle this as special case because cbind for empty tasks is broken
-        return(task)
+        return(task$select(character(0)))
```
mb706, line 102: what is this doing?

mb706:
Excellent work

PR #359, opened 2020-03-04
Fix imputation with missings

pfistfl:
This should fix https://stackoverflow.com/questions/60512348/how-to-impute-data-with-mlr3-and-predict-with-na-values

Problem was:
`PipeOpMissInd`, when trained on a task without missing data returned the full (unchanged) task during prediction.
This led to a situation where the ordere of `PipeOp's` in `gunion` mattered, as the second task would overwrite the first one. 

This patch returns an empty task, in case no indicators are added during training.

There should be some more info that columns are overwritten in `PipeOpFeatureUnion`, will open an issue for this.


pfistfl:
Closing here, succeeded by #358

PR #361, opened 2020-03-08
Branching ppl sugar

pfistfl:
This would allow to do branch - unbranch operations as follows:

```r
lrns = map(list(lrn("classif.rpart"), lrn("classif.featureless")), po)
gr = ppl("branch", lrns)
```

Currently, the automatic setting of dependencies leads to errors if  `default` values are specified.


pfistfl:
Bloacked by https://github.com/mlr-org/paradox/issues/240 and https://github.com/mlr-org/paradox/issues/216

pfistfl:
We do deps when constructing the graph, this can be merged then.

PR #366, opened 2020-03-11
Pipeop imputemode

### LINKED ISSUE #312: Missing PipeOp: PipeOpImputeMode
Could call `mlr3misc::compute_mode()`.

sumny:
Closes  #312 
Regarding tests, I just added some tests in the `PipeOpImpute` bogus `PipeOp` that do the same as for mean or median imputing. However see:
https://github.com/mlr-org/mlr3pipelines/blob/3fe02235ccdfd74ee4713cde39031bce24d17f4e/tests/testthat/test_pipeop_impute.R#L5 (comment by @mb706 )
So we may want to clean up the test in general.


R/PipeOpImputeMode.R
```diff
@@ -0,0 +1,88 @@
...
+      ps$values = list(ties_method = "random")
+      super$initialize(id, param_set = ps, param_vals = param_vals, packages = "mlr3misc")
+    },
+
+    select_cols = function(task) task$feature_types[get("type") %in% c("numeric", "integer"), get("id")],
+
+    train_imputer = function(feature, type, context) {
+      mod = mlr3misc::compute_mode(feature, ties_method = self$param_set$values$ties_method)
+      if (is.na(mod)) {
+        mod = 0
```
mllg: 1. Cannot be NA unless `na_rm = TRUE` is passed to `compute_mode()`
2. The mode should be one of the observed values, imputing 0 is not reasonable.
3. It is not clear what value to impute if all values of an integer are NA. What are we doing here while imputing with the mean? -- for factors, we could just sample a random factor level.

R/PipeOpImputeMode.R
```diff
@@ -0,0 +1,88 @@
+#' @title PipeOpImputeMode
+#'
+#' @usage NULL
+#' @name mlr_pipeops_imputemode
+#' @format [`R6Class`] object inheriting from [`PipeOpImpute`]/[`PipeOp`].
+#'
+#' @description
+#' Impute numerical features by their mode.
```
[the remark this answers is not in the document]
sumny: makes total sense.. fixed

R/PipeOpImputeMode.R
sumny: 1. fixed, thanks.

2. and 3. `PipeOpMean` and `PipeOpMedian` both always impute `0` for numerics or integers if all are `NA`. For `factor`s (also `ordered`s) `PipeOpImputeMode` now samples a level randomly. (Does it make sense to modify the behavior here to also match the `ties_method` argument? I.e. if all values are `NA`, then either a random level is sampled - `random`, or the `first` or `last` level is choosen). Logicals are now also supported (if all values are `NA` `TRUE`/`FALSE` is being sampled).

mllg: > Logicals are now also supported (if all values are `NA` `TRUE`/`FALSE` is being sampled).

Great!


>   `PipeOpMean` and `PipeOpMedian` both always impute `0` for numerics or integers if all are `NA`

Okay ... I find it odd, but at least it is properly documented. Then we should do the same thing here.

> Does it make sense to modify the behavior here to also match the ties_method argument? I.e. if all values are NA, then either a random level is sampled - random, or the first or last level is choosen

Yes, makes sense.

sumny: > Okay ... I find it odd, but at least it is properly documented. Then we should do the same thing here.

This also the default for `PipeOpImputeMode` now and it is also documented.

> Yes, makes sense.

Okay, fixed this. Maybe @mb706 wants to leave a final comment regarding the tests!

mb706:
I want to look at the mode length == 0 case

PR #371, opened 2020-03-30
use only mlr3 task_types to set the task and prediction types

### LINKED ISSUE #368: Error creating pipeop learner when using mlr3spatiotempcv package

sumny:
additional packages, e.g., `mlr3spatiotempcv` add their own `task_type`s to `mlr_reflections` which conflicts when constructing a `PipeOpLearner` after the package is loaded. Therefore, fall back to the `mlr3` `task_type`s in `mlr_reflections`. Closes #368


sumny:
This probably needs more discussion because this most likely is a unique interaction with `mlr3spatiotempcv` only and maybe the fix should go there https://github.com/mlr-org/mlr3spatiotempcv/blob/e5332e4c64898248c51515ee47bb514c47cb331e/R/zzz.R#L18

pfistfl:
After thinking about this a little more, I guess the correct thing would be to sort the table, so that mlr3 always comes first and then just always only using the first element. 
This would not disable things working for e.g. `mlr3proba`

pfistfl:
Summarizing what I recall from talking to Bernd: 
- Finish this up, use the approach I described above. 

Martin / I created issues for this rest of the problem in `mlr3`, that maybe can make this more beautiful in the futrue.

R/PipeOpLearner.R
```diff
@@ -80,8 +80,10 @@ PipeOpLearner = R6Class("PipeOpLearner", inherit = PipeOp,
   public = list(
     initialize = function(learner, id = if (is.character(learner)) learner else learner$id, param_vals = list()) {
       private$.learner = as_learner(learner)$clone(deep = TRUE)  # FIXME: use `clone=TRUE` when mlr-org/mlr3#344 is fixed
-      task_type = mlr_reflections$task_types[private$.learner$task_type]$task
-      out_type = mlr_reflections$task_types[private$.learner$task_type]$prediction
+      task_types = copy(mlr_reflections$task_types)
+      setorder(task_types, package)
+      task_type = task_types[type == private$.learner$task_type]$task[1L]
```
mb706, line 85: does
```r
task_types[type == private$.learner$task_type][order(package)[1], "task"]
```
also work here? (or something like this, I can't try it rn, I hope you get what I mean) if so it would be nice to avoid the `copy` and `setorder`.

also: please add a #fixme comment with a link to the relevant issue in mlr3spatiotemp and / or mlr3

also: I think you should use `get("type")` if you refer to variables *inside* the `data.table` because the static checker of cran will complain otherwise.

sumny: good suggestions! Thanks! I dropped the copy and setorder and added a FIXME linking to mlr-org/mlr3#470

PR #374, opened 2020-03-31
Pipeop learner and learnercv if a Graph and no id is used, collapse the ids of the Graph

### LINKED ISSUE #292: learner_cv construction fails if given a Graph and no id

sumny:
Closes #292


R/PipeOpLearner.R
```diff
@@ -79,6 +79,12 @@
 PipeOpLearner = R6Class("PipeOpLearner", inherit = PipeOp,
   public = list(
     initialize = function(learner, id = if (is.character(learner)) learner else learner$id, param_vals = list()) {
```
mb706, line 81: I think the solution here would be to have `id = NULL`, and then in the code
```r
private$.learner = as_learner(learner, clone = TRUE)  # merge master before doing this!
self$id = id %??% private$.learner$id
```
also add a tests pls

Issue #364, opened 2020-03-10
Design for inverter pipeop


mb706:
Just trying to write this down as fast as possible, sorry if this is a bit incoherent.

We want to have pipeops that somehow communicate with some part later in the graph, example is target trafos, that have to be inverted. It may help to look up `TOCPO` in mlrCPO doc to see what target trafo and inversion usually entails. In short: You need to transform the target during the training phase, and you need to invert the prediction done in the prediction phase. This inversion may need info on both the training data *and* the prediction data. (Imagine a trafo that subtracts the value of feature 1 from the target for each sample. Then the inversion phase might want to add the value of feature 1 from each corresponding prediction sample).

We want to be able to concentrate all the interesting logic in a single place, so we will use the following design, which was an idea of @pfistfl :
```
 -- [ trafo ] ----- [ learner ] ----- [inverter] -->
       \                                  /
        ------------{ function() } ------
```
the `trafo` transforms the target and also sends a function down the line which the inverter applies.

In the following it is necessary to indicate the "types" of pipeop channels. I will use `[<train>,<predict>]` for that. So the input channel of a `PipeOpLearner` has type `[Task,Task]`, the output channel has `[NULL,Prediction]`. `PipeOpNop` has `[*,*]` for its input and output channel.

We will have a base class for the inverter which we will call `PipeOpFunctionApplicator` for now. It has two input channels. One `"fun"` (I am avoiding calling it `function` because that name clashes with the keyword) type `[function,function]`, and one `"..."` (i.e. a vararg channel) of type `[*,*]`. It has one output channel `[*,*]`. Both in train and predict (_internal) this should apply the "fun" function to its input channels; it will look something like `function(inputs) { list(inputs[[1]](inputs[-1])) }`. The function that comes in through `"fun"` therefore must take one argument which will be a list. (Note the train function must set the `$state`)

We will inherit from this as `PipeOpInverter`. Notably it does not have a vararg input, it only has two channels. Both channels take `NULL` during training, because that is what `PipeOpLearner` generates during training. The input channel types are thus `"fun"` `[NULL,function]`, `"prediction"` `[NULL,Prediction]`. The output channel type is `[NULL,Prediction]`. The `train_internal` function is just `state = list() ; NULL`, the `predict_internal` function is inherited from `PipeOpFunctionApplicator`.

[...]

Issue #322, opened 2020-02-12
PipeOp for target preprocessing


mb706:
We want to do things to the task that can be inverted later. This can be simple target operations ([see TOCPO](https://cran.r-project.org/web/packages/mlrCPO/vignettes/z_4_custom_CPOs_terse.html#makecpotargetop)), or more complicated stuff. The issues we have that would profit from this are  https://github.com/mlr-org/mlr3pipelines/issues/183, https://github.com/mlr-org/mlr3pipelines/issues/189, maybe https://github.com/mlr-org/mlr3pipelines/issues/301(?). @pfistfl idea was to write a single generic inverter PipeOp that does something to a Prediction / to multiple Predictions, and that whatever happens is determined by an inverter function that is sent along a channel.

We should probably write a class that facilitates easy generation of the inverter function. The user should overload a train(), a predict(), and an invert() function, and the class takes care of generating the function that is sent along the edge to the inverter PipeOp. This is similar to how PipeOpTaskPreproc operates, and maybe this can even be tacked on to PipeOpTaskPreproc. The only issue here is that PipeOpTaskPreproc is for a single task output, while our TOCPO may want to generate multiple tasks that are then put back together by the inverter.

PipeOps to write:
* [x] PipeOpInvert
* [x] PipeOpInvertiblePreproc. Or maybe put this together with PipeOpTaskPreproc, not sure.

sumny:
#364

pfistfl:
po_ytrafo %>>% ... graph  .... %>>% po_ytrafoinverter
          \ -----------------------------------------------/

po_ytrafo = Abstract;
Slots:
 trafo():  e.g. log
invert(): e.g. exp
state (e.g. scaling factors)

po_ytrafo_log; sets trafo = log; invert = exp

pfistfl:
Same as #364, closing here.

PR #376, opened 2020-03-31
ytrafo pipeops

sumny:
See #322 and #364 
This is still WIP.
Adds abstract `PipeOpInvertiblePreproc` from which you can inherit your own trafo `PipeOp`. Toy example is given in `PipeOpSimpleTrafo`. Together with `PipeOpInverter` this allows to build a `Graph` with a `Learner` in the middle, so that the target is transformed during training and retransformed for prediction. Docu is drafted; tests missing. Let's discuss this first


R/PipeOpTrafo.R
```diff
@@ -0,0 +1,240 @@
...
+    initialize = function(id = "logtrafo") {
+      super$initialize(id = id)
+    },
+
+    train_target = function(task) {
+      self$state = list()
+      new_target = log(task$data(cols = task$target_names))
+      names(new_target) = paste0("log.", task$target_names)
+      task$cbind(new_target)
+      private$update_target(task, new_target = names(new_target), new_type = NULL)
```
mb706: how about a signature
```r
private$update_target(task, new_target, new_target_name = NULL, new_type = NULL)
```
and calling it here as
```r
new_target = log(task$data(cols = task$target_names))
private$update_target(task, new_target = new_target,
  new_target_name = paste0("log.", task$target_names))
```

pfistfl:
Please test that you can construct an arbitrary `log(base = X)` -> invert = `x ^ base` trafo.

R/PipeOpTrafo.R
```diff
@@ -0,0 +1,240 @@
...
+
+    inverter = function(prediction, predict_phase_control) {
+      # function that inverts the predictions and returns a Prediction object
+      stop("Abstract.")
+    }
+  ),
+  private = list(
+    update_target = function(task, new_target, new_type = NULL, ...) {
+      type = if (is.null(new_type)) task$task_type else assert_subset(new_type, mlr_reflections$task_types$type, empty.ok = FALSE)
+      get(mlr_reflections$task_types[type]$task)$new(id = task$id, backend = task$backend, target = new_target, ...)
```
pfistfl: See `convert_target` from hurdle models branch;  do not convert task if task_type stays the same.

R/PipeOpTrafo.R
```diff
@@ -0,0 +1,240 @@
...
+
+    invert_help = function(predict_phase_control) {
+      function(inputs) {
+        assert_list(inputs, len = 1L, types = "Prediction")
+        self$inverter(inputs[[1L]], predict_phase_control)
+      }
+    },
+
+    set_target_to_NA_and_possibly_convert_task_type = function(task) {
+      #FIXME: convert the task target to NA and convert the task type?
```
pfistfl: I guess this should just set the  `col_roles` of the original target to 0.

R/PipeOpTrafo.R
```diff
@@ -0,0 +1,240 @@
...
+
+    predict_internal = function(inputs) {
+      list(inputs[[1L]](inputs[-1L]))
+    }
+  )
+)
+
+mlr_pipeops$add("inverter", PipeOpInverter)
+
+#' @title PipeOpLogtrafo
```
pfistfl: PipeOpSimpleTrafo:
- following params in its `param_set`: `trafo`, `inverter`, `new_target_name`, `new_task_type`
- `trafo` and `inverter` default to `identity`
- `new_target_name` defaults to `paste0("trafo_", xxold_target_name)`
- `new_task_type` defaults to `NULL` (keep old task type)

pfistfl: then `PipeOpLogTrafo` could inherit from this and just define different `trafo` and `inverter`.

sumny: done, maybe we discuss whether another inheritance from `PipeOpSimpleTrafo` is needed for `PipeOpLogTrafo` (it is pretty easy to set this up already in `PipeOpSimpleTrafo`)

sumny:
> Please test that you can construct an arbitrary `log(base = X)` -> invert = `x ^ base` trafo.

Added an example in `PipeOpSimpleTrafo`.

pfistfl:
Could you also add the `ppl` shortcut and also document this?

R/PipeOpNewTarget.R
```diff
@@ -0,0 +1,43 @@
+#' Convert a task from its type to another.
+#'
+#' The task's target is replaced by a different column from the data.
+#' #FIXME: This needs detailed tests.
```
pfistfl, line 4: + merge relevant tests from the old PR.

R/PipeOpNewTarget.R
```diff
@@ -0,0 +1,43 @@
...
+#'   A [`Task`][mlr3::Task] to be converted.
+#' @param new_target `character(1)`\cr
+#'   New target to be set, must be a column in the `intask` data.
+#'   If `NULL`, no new target is set, and task is converted as-is.
+#' @param new_type `character(1)`\cr
+#'   The new task type. Must be in `mlr_reflections$task_types`.
+#'   If `NULL` (default), a new task with the same task_type is created.
+#' @param drop_original_target `logical(1)`\cr
+#'   If `FALSE` (default), the original target is added as a feature.
+#'   Otherwise the original target is assigned no col_role, i.e. dropped.
```
pfistfl, line 15: We assign `character(0)` in order to drop col_roles

R/PipeOpTrafo.R
```diff
@@ -0,0 +1,348 @@
...
+#'po$predict(list(task))
+#'
+#'g = Graph$new()
+#'g$add_pipeop(po)
+#'g$add_pipeop(LearnerRegrRpart$new())
+#'g$add_pipeop(PipeOpInverter$new())
+#'g$add_edge(src_id = "logtrafo", dst_id = "inverter", src_channel = 1, dst_channel = 1)
+#'g$add_edge(src_id = "logtrafo", dst_id = "regr.rpart", src_channel = 2, dst_channel = 1)
+#'g$add_edge(src_id = "regr.rpart", dst_id = "inverter", src_channel = 1, dst_channel = 2)
+#'
```
pfistfl, line 305: Also show the `ppl("targettrafo", ...)` version here.

sumny:
Renamed the `PipeOp`s, added tests, and also a `ppl` that allows for an easier build around a `PipeOpLearner` or `Graph` (where the last `PipeOp` is a `PipeOpLearner`). Also I am more strict regarding the `Task` and `Prediction` within `PipeOpTargetTrafoSimple` - I think this is currently only useful for regression tasks where the target is a (continous) numerical variable (but you can prove me wrong). This is now ready for a final review!

R/PipeOpTrafo.R
```diff
@@ -0,0 +1,365 @@
...
+    initialize = function(id, param_set = ParamSet$new(), param_vals = list(), packages = character(0), tags = NULL) {
+      super$initialize(id = id, param_set = param_set, param_vals = param_vals,
+        input = data.table(name = "input", train = "Task", predict = "Task"),
+        output = data.table(name = c("fun", "output"), train = c("NULL", "Task"), predict = c("function", "Task")),
+        packages = packages, tags = tags
+      )
+    },
+
+    train_target = function(task) {
+      # return a modified task and set the state
```
pfistfl, line 109: do we still need all the comments?

R/PipeOpTrafo.R
```diff
@@ -0,0 +1,365 @@
...
+#'
+#' @section Internals:
+#' Overloads [`PipeOpTargetTrafo`]'s `train_target()`, `train_invert()` and `inverter()`.
+#' Should be used in combination with [`PipeOpTargetInverter`].
+#'
+#' @section Methods:
+#' Only methods inherited from [`PipeOpTargetTrafo`]/[`PipeOp`].
+#'
+#' @examples
+#'library(mlr3)
```
pfistfl, line 291: indent here

R/mlr_graphs_elements.R
```diff
@@ -243,3 +244,67 @@ pipeline_branch = function(graphs, prefix_branchops = "", prefix_paths = FALSE)
 }
 
 mlr_graphs$add("branch", pipeline_branch)
+
+
+
+#' @title Transform A Numerical Target And Later Invert The Transformation
```
pfistfl, line 250: Transform and re-transform the target variable

R/mlr_graphs_elements.R
```diff
@@ -243,3 +244,67 @@ pipeline_branch = function(graphs, prefix_branchops = "", prefix_paths = FALSE)
...
+#' @name mlr_graphs_targettrafo
+#' @description
+#' Creates a [`Graph`] that transforms a numerical target of a [`TaskRegr`][mlr3::TaskRegr] during
```
pfistfl, line 253: I would drop the numerical, this works for any target.

pfistfl, line 253: Wraps a graph

R/mlr_graphs_elements.R
```diff
@@ -243,3 +244,67 @@ pipeline_branch = function(graphs, prefix_branchops = "", prefix_paths = FALSE)
...
+#')
+#'g$add_pipeop(LearnerRegrRpart$new())
+#'g$add_pipeop(PipeOpTargetInverter$new())
+#'g$add_edge(src_id = "targettrafosimple", dst_id = "targetinverter",
+#'  src_channel = 1, dst_channel = 1)
+#'g$add_edge(src_id = "targettrafosimple", dst_id = "regr.rpart",
+#'  src_channel = 2, dst_channel = 1)
+#'g$add_edge(src_id = "regr.rpart", dst_id = "targetinverter",
+#'  src_channel = 1, dst_channel = 2)
+pipeline_targettrafo = function(graph) {
```
pfistfl, line 291: signature should be `function(graph, trafo_pipeop = NULL, id_prefix = "")`
If  `trafo_pipeop = NULL` uses `PipeOpTargetTrafoSimple`, otherwise a custom target trafo can be provided.
Additionally prefixes the `id` of trafo and inverter with `id_prefix`. 
Otherwise, this could lead to issues as duplicate id's are not allowed.

Test that this works for custom id's aswell.

R/PipeOpNewTarget.R
```diff
@@ -0,0 +1,43 @@
+#' Convert a task from its type to another.
```
pfistfl, line 1: File Name is wrong

tests/testthat/test_pipeop_newtarget.R
```diff
@@ -0,0 +1,20 @@
+test_that("convert_task reconstructs task", {
```
pfistfl, line 1: could you copy all relevant tests?  (Or possibly test a few more edge cases).

In theory, this function should also be able to just change iris into a regr task by predicting `Sepal.Length` instead of `Species` etc. guess we  can do a few more sensible tests.

tests/testthat/test_pipeop_targettrafosimple.R
```diff
@@ -0,0 +1,74 @@
+context("PipeOpTargetTrafoSimple")
```
pfistfl, line 1: So the following currently breaks, as it would require a `PredictionClassif` -> `PredictionRegr` conversion
- Task: `boston_housing`, `trafo = x > 20`,  inverter if (x) 19 else 21 (then we'd need a classif learner)

While the following might work. This is actually somewhat relevant.
- Task: `boston_housing`,`trafo = x > 20`,  inverter `identity`? This should probably be possible.

pfistfl: Could we test that and possibly add a fixme that we might somehow enable the first option in the future?

tests/testthat/test_pipeop_targettrafosimple.R
```diff
@@ -0,0 +1,74 @@
...
+ dat$medv = log(dat$medv, base = 2)
+ task_log = TaskRegr$new("boston_housing_log", backend = dat, target = "medv")
+
+ learner = LearnerRegrRpart$new()
+ learner$train(task_log)
+
+ learner_predict_out = learner$predict(task_log)
+ expect_equal(2 ^ learner_predict_out$truth, predict_out[[1L]]$truth)
+ expect_equal(2 ^ learner_predict_out$response, predict_out[[1L]]$response)
+})
```
pfistfl, line 74: We currently have no checks for the state, i.e. whether learning and applying a state possibly works.

I guess we could do one `PipeOp` that covers this in the PR's comments.

pfistfl:
We currently can not test that the `state` actually works as we have no `PipeOp` that uses it.

I will describe a `PipeOp` we should implement here in order to showcase how such pipeops are implemented and to test that this works in general.
```
PipeOpTargetTrafoScaleRange
...
basically does the same as `PipeOpScaleRange`, i.e. stores range trafo to state.
additionally, it knows how to retransform scaled data.
```

tests/testthat/test_pipeop_targettrafosimple.R
sumny: yes, seems like a good idea!

tests/testthat/test_pipeop_targettrafosimple.R
sumny: I added two more tests here, one that goes from  `PredictionClassif` -> `PredictionRegr` and one the other way. Both are not the most meaningful ones but it is possible. For most cases to actually be meaningful the user will rely on setting the `$state` and rely on this for the inversion (like statet by you below).

sumny:
> We currently can not test that the `state` actually works as we have no `PipeOp` that uses it.
> 
> I will describe a `PipeOp` we should implement here in order to showcase how such pipeops are implemented and to test that this works in general.
> 
> ```
> PipeOpTargetTrafoScaleRange
> ...
> basically does the same as `PipeOpScaleRange`, i.e. stores range trafo to state.
> additionally, it knows how to retransform scaled data.
> ```

I added a `PipeOpTargetTrafoScaleRange` and tests. Seems to work nice. Had to add a `set_state()` method to `PipeOpTargetTrafo` that can be overloaded that makes such trafos way easier. `set_state()` is called a single time in `.train()`, before `train_task()` is evaluated so you can then rely on the `$state` both during training and prediction.

PR #382, opened 2020-04-01
Caching Prototype

pfistfl:
# Caching

These docs describe `oportunistic caching`, i.e. caching after a first function call.
If the same function is executed twice in parallel, this does not save any time/cores.
The example currently uses the `R.cache` package by Henrik Bengtsson for caching.
This is just a very simple caching package, that provides a clean, simple API, could
theoretically be replaced by other packages.

## Implementation Details

Ideally we would like to do caching on an abstract level,
perhaps within the PipeOp base-classes `$train` function.
A very nice point would be to wrap the call to `private$.train`.
This would make complexity very manageable.

`R.cache::evalWithMemoization` memoizes the provided expression.
The `hash` is computed from its `key` argument.

Possible solution: adjust `PipeOp` in `PipeOp.R`
```
train = function(input) {
      ...
      t = check_types(self, input, "input", "train")
      # caching
      R.cache::evalWithMemoization({
        result = list(private$.train(input), self$state) #(see A below)
        }, key = list(map_chr(input, get_hash), self$hash)
      )
      if (is.null(self$state)) state = result$state #(see A below)
      output = check_types(self, result$output, "output", "train")
      output
    },
    predict = function(input) {
      ...
      input = check_types(self, input, "input", "predict")
      R.cache::evalWithMemoization({
        output = private$.predict(input)
        },
         key = list(map_chr(input, get_hash), self$hash)
      )
      output = check_types(self, output, "output", "predict")
      output
    }
  ),
```

where `get_hash` is:
```
get_hash = function(x) {
  if (!is.null(x$hash)) return(x$hash)
    digest(x, algo = "xxhash64")
}
```

## Possible problems:

A) Unfortunately `private$.train()` is not a pure function, but
   instead has side-effects:
    - sets a `$state`
    - ... (others?)

If we can ensure that the only side-effect of `$.train` is a modified state, 
we could also memoize the state during `$train` (see above).

If other fields are updated, we need to have a list of fields that are updated or go a different route.

## Further Issues:

F) Should caching be optional? 
   Probably yes!

G) How do we globally enable/disable caching?
    1. global option
    < ugly, might not work with parallelization. >

    2. caching can be turned on in `Graph` | `GraphLearner`
    ```
    Graph = R6Class(
     ...
     caching = TRUE,
     ...
    )
    ```
    `GraphLearner` gets an active binding to turn caching of it's graph on/off.
    Could also be added as an arg to the `GraphLearner`s constructor.

    The caching of individual steps is then done by adjusting calls to `graph_reduce`:
    `graph_reduce(..., caching = self$caching)`

H) Should caching be disabled for some `PipeOp`s?
   Yes, possible solution: New field in each `PipeOp`: `cached`.
   Caching for a pipeop only happens if `cached = TRUE`.
   Can also be manually changed to disable caching for any pipeop.

Open Questions:
  - How do `$train` and `$predict` know whether to do caching or not?
    Add a second argument `caching`?
  - How do caching and `parallelization` interact?
  - Does `R.cache::evalWithMemoization`s `key` arg need anything else?

### DIFF
--- a/R/PipeOp.R
+++ b/R/PipeOp.R
@@ -241,14 +241,13 @@ PipeOp = R6Class("PipeOp",
         return(named_list(self$output$name, NO_OP))
       }
       input = check_types(self, input, "input", "train")
-      keya <<- list(map_chr(input, get_hash), self$hash)
+      # caching
       R.cache::evalWithMemoization({
-        output = private$.train(input)
-        },
-        key = list(map_chr(input, get_hash), self$hash)
+        result = list(private$.train(input), self$state)
+        }, key = list(map_chr(input, get_hash), self$hash)
       )
-      keyb <<- list(map_chr(input, get_hash), self$hash)
-      output = check_types(self, output, "output", "train")
+      if (is.null(self$state)) state = result$state
+      output = check_types(self, result$output, "output", "train")
       output
     },
     predict = function(input) {
@@ -262,14 +261,13 @@ PipeOp = R6Class("PipeOp",
         stopf("Pipeop %s got NO_OP during train but no NO_OP during predict.", self$id)
       }
       input = check_types(self, input, "input", "predict")
-
       R.cache::evalWithMemoization({
         output = private$.predict(input)
         },
          key = list(map_chr(input, get_hash), self$hash)
       )
-      print(digest(list(class(self), self$id), algo = "xxhash64"))
-      print(digest(self$param_set, algo = "xxhash64"))
+      # print(digest(list(class(self), self$id), algo = "xxhash64"))
+      # print(digest(self$param_set, algo = "xxhash64"))
       output = check_types(self, output, "output", "predict")
       output
     }
--- a/R/PipeOp.R
+++ b/R/PipeOp.R
@@ -241,7 +241,17 @@ PipeOp = R6Class("PipeOp",
         return(named_list(self$output$name, NO_OP))
       }
       input = check_types(self, input, "input", "train")
-      output = private$.train(input)
+      print(self$hash)
+      print(digest(private$.param_set, algo = "xxhash64"))
+      a <<- private$.param_set
+      R.cache::evalWithMemoization({
+        output = private$.train(input)
+        },
+        key = list(map_chr(input, get_hash), self$hash)
+      )
+      print(self$hash)
+      print(digest(private$.param_set, algo = "xxhash64"))
+      b <<- private$.param_set
       output = check_types(self, output, "output", "train")
       output
     },
@@ -255,9 +265,15 @@ PipeOp = R6Class("PipeOp",
       if (is_noop(self$state)) {
         stopf("Pipeop %s got NO_OP during train but no NO_OP during predict.", self$id)
       }
-
       input = check_types(self, input, "input", "predict")
-      output = private$.predict(input)
+
+      R.cache::evalWithMemoization({
+        output = private$.predict(input)
+        },
+         key = list(map_chr(input, get_hash), self$hash)
+      )
+      print(digest(list(class(self), self$id), algo = "xxhash64"))
+      print(digest(self$param_set, algo = "xxhash64"))
       output = check_types(self, output, "output", "predict")
       output
     }
@@ -365,4 +381,9 @@ check_types = function(self, data, direction, operation) {
   }
   names(data) = typetable$name
   data
+}
+
+get_hash = function(x) {
+  if (!is.null(x$hash)) return(x$hash)
+    digest(x, algo = "xxhash64")
 }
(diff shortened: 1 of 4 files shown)


mb706:
My thoughts about things to consider, in random order:
* with some operations it may make more sense to save just the `$state` and not the result. Then during `$train()` the caching mechanism can set the state from cache and call `$.predict()`.
* `PipeOp`s should contain metadata about whether they are deterministic or not, and whether their `.train()` and `.predict()` results are the same whenever the input to both is the same (use common vs. separate cache)
* caching in `mlrCPO` was a wrapper-PipeOp, we could also have that here. Pro: For multiple operations only the last output needs to be saved; makes the configuration of different caching mechanisms easier. Cons: We get the drawbacks of wrapping: the graph structure gets obscured. Also when wrapping multiple operations and just one of them is nondeterministic everything falls apart. We may want a `ppl()` function that wraps a graph optimally so that linear deterministic segments are cached together and only the output of the last `PipeOp` is kept. (Also works for arbitrary `Graph`s).

pfistfl:
I added your comments and some responses.

pat-s:
Have to we concluded to do this in a per-package base now rather than upstream in mlr3?

pfistfl:
> Have to we concluded to do this in a per-package base now rather than upstream in mlr3?

This would actually be broader then doing it in `mlr3`.
As every step in the pipeline is potentially cached, this includes
- `learners`
- `filters`
- `data transform pipeops`

The only drawback would then be that in order to benefit from caching, those would need to be part of a pipeline, which they should be anyways in most cases.

`benchmark()` would then be cached by wrapping each `learner` inside a `GraphLearner`,
 i.e.
`lrns = map(lrns, function(lrn) GraphLearner$new(po(lrn)))`

pat-s:
OK. I again want to raise awareness that people who want to use mlr3 without pipelines should also profit from caching. For example, when filtering users should be able to profit from caching but this would require adding a per-package caching implementation then (there might be more ext packages besides filters). This one could potentially conflict with the pipelines caching.

Also, did you look how [drake](https://github.com/ropensci/drake) does this? In the end its also a workflow package that aims to cache steps within the "Graph" and detect if there are steps in the cache that do not need to be rerun.
Just trying to potentially save your time for tasks that might reinvent the wheel - even though mlr3pipelines might be completely different to how drake does it.

pfistfl:
>  This one could potentially conflict with the pipelines caching.

In general, different levels of caching should not interfere, the worst case I could imagine would be to cache things twice, i.e. `PipeOp` and `Filter` cache their results. This would just mean that `PipeOp`s that operate on something that itself knows how to cache things would be adjusted to deactivate lower-level caching.

> people who want to use mlr3 without pipelines should also profit from caching.

I can not really judge this, but I am not sure I agree. I agree that we should provide enough convenience functions to enable people to work without learning all the in's and out's of 'mlr3pipelines`

Currently it is as complicated as
```r 
flt("variance") %>>% lrn("classif.rpart")
```
 to write a filtered learner.

So correct me if I am wrong, but when would you do filtering without using `mlr3pipelines`? 
That would only be the case when you do train/test split and filtering manually?

I have not looked at `drake` too deep but the current caching implementation that covers everything `mlr3pipelnes` needs has < 40 lines.

pat-s:
> This would just mean that PipeOps that operate on something that itself knows how to cache things would be adjusted to deactivate lower-level caching.

Yeah or maybe to rely on the lower-level caching implementation if it exists.

> So correct me if I am wrong, but when would you do filtering without using mlr3pipelines?
> That would only be the case when you do train/test split and filtering manually?

Yeah maybe it does not make sense and mlr3pipelines is a "must-use" in the whole game. For now I've only done toy benchmarks without any wrapped learners - all of these are still written in the old mlr.
And I can probably also combine drake and mlr3pipelines, maybe even both caching approaches.

Maybe you find [mlr3-learnerdrake](https://github.com/mlr-org/mlr3-learndrake) interesting. We/I should extend it with a pipelines example.

pfistfl:
Review Bernd:
- Should PipeOp Ids be cached
- Release this in a separate release
- Add blogpost, showcase results
- Thoroughly testride

R/PipeOp.R
```diff
@@ -296,6 +310,26 @@ PipeOp = R6Class("PipeOp",
...
+    },
+    cache_state = function(val) {
+      if (!missing(val)) {
+        stop("cache_state is read-only!")
+      } 
+      private$.cache_state
+    },
+    stochastic = function(val) {
+      if (!missing(val)) {
+        private$.stochastic = assert_subset(val, c("train", "predict"))
```
mb706, line 329: shouldn't this be read-only and set during initialization?

Timo-Ko:
Hi folks, we have a use case in mlr3 where caching inside mlr3pipelines would be super useful:
We have a graph learner that first imputes missing values and then an auto-tuned RF: lrn_rf_po = po("imputeoor")  %>>% rf_tuned
Now, we would like to access the imputed values from our pipe.
Do you happen to have any updates / working code for the caching?
Thanks a lot in advance!

Issue #306, opened 2020-01-25
predict_type from last learner not propagated


berndbischl:
```
library(mlr3)
library(mlr3pipelines)

task = tsk("pima")
learner = lrn("classif.rpart", predict_type = "prob")
resampling = rsmp("holdout")
r = resample(task, learner, resampling)
m = msr("classif.auc")
r$aggregate(m)  # works

pipeline = po("imputemean") %>>% learner
r = resample(task, pipeline, resampling)
r$aggregate(m) # fails
# Error: Measure 'classif.auc' needs predict_type 'prob'

pipeline = po("imputemean") %>>% learner
graphl = GraphLearner$new(pipeline)
graphl$predict_type = "prob" # works but looks tedious
r = resample(task, graphl, resampling)
r$aggregate(m)
```

I ran thru the usecase above. 
1) Where is exactly documented what the line 
graphl$predict_type = "prob" 
does or how predict types in general work for pipelines

2) for my usecase above it seems pretty unintuitive for the user why he has to set the predict_tyoe again? can we simplify this?

mb706:
1. this is not documented explicitly; I guess it is implicit in that (a) a Learner needs a predict_type, which is initialized as "response", (b) every GraphLearner is a Learner (Aristotle would be proud). But yeah should be written somewhere.

2. Unfortunately we can only solve this heuristically, because the last operation may not be a Learner at all (and may in fact be a thresholding pipeop that could change "response" to "prob" predictions). A few solutions I see:

    1. GraphLearner always inspects the last PipeOp; if it is a PipeOpLearner it checks it for a predict_type, otherwise it defaults to "response"
    2. The PipeOp gets a predict_type, so a PipeOpLearner gets its predict_type from the attached Learner, but a PipeOpThreshold can also get a predict_type that may be different
        1. Maybe we give PipeOps an additional predict_type "pass-through" that signals that the previous PipeOp's predict_type should be used. E.g. if the last PipeOp is a prediction averager that maps prob -> prob and response -> response we could cover that case

    What should the behaviour be when a user *changes* `graph$predict_type`, vs. `graph$pipeops$classif.rpart$learner$predict_type`?

pfistfl:
I propose the following design:

* `predict_type` is an AB in every `PipeOp` that returns NULL for almost every `PipeOp`.
* Any PipeOp that changes the `predict_type` has an active binding that reflects this, i.e.
[...]

PR #394, opened 2020-04-05
Predict type set get

mb706:


R/GraphLearner.R
```diff
@@ -102,6 +94,37 @@ GraphLearner = R6Class("GraphLearner", inherit = Learner,
...
+    },
+    get_predict_type = function() {
+      # recursively walk backwards through the graph 
+      get_po_predict_type = function(x) {
+        if (!is.null(x$predict_type)) return(x$predict_type)
+        prdcssrs = self$graph$edges[dst_id == x$id, ]$src_id
+        if (length(prdcssrs)) {
+          # all non-null elements
+          predict_types = keep(map(self$graph$pipeops[prdcssrs], get_po_predict_type), Negate(is.null))
+          if (length(unique(predict_types)) == 1L)
```
mb706, line 106: its a bit weird to ignore NULL-returns by `get_po_predict_type` calls, but *also* encode inconsistent predecessor predict types as `NULL`.

R/GraphLearner.R
```diff
@@ -102,6 +94,37 @@ GraphLearner = R6Class("GraphLearner", inherit = Learner,
         return(value$clone(deep = TRUE))
       }
       value
+    },
+    get_predict_type = function() {
+      # recursively walk backwards through the graph 
+      get_po_predict_type = function(x) {
+        if (!is.null(x$predict_type)) return(x$predict_type)
```
mb706, line 101: `x$predict_type` should go into the base class

R/GraphLearner.R
```diff
@@ -102,6 +94,37 @@ GraphLearner = R6Class("GraphLearner", inherit = Learner,
         return(value$clone(deep = TRUE))
       }
       value
+    },
+    get_predict_type = function() {
```
mb706, line 98: comment smth like "trace every path backward from the sink until a `x$predict_type` is found for each of these paths (or there are no predecessors)"

R/PipeOpLearner.R
```diff
@@ -118,6 +119,13 @@ PipeOpLearner = R6Class("PipeOpLearner", inherit = PipeOp,
         }
       }
       private$.learner
+    },
+    predict_type = function(val) {
+      if (!missing(val)) {
+        assert_subset(val, names(mlr_reflections$learner_predict_types[[private$.learner$task_type]]))
```
mb706, line 125: doesn't the learner itself also assert this?

pfistfl: pipeops can / could have AB's that are not learner's. But those might assert this themselves aswell.

pfistfl: we should think about what PipeOpThreshold does in this case. There we can not set to `prob`.
this should then possibly error (within PipeOpThreshold I guess).

R/GraphLearner.R
pfistfl: true

mb706:
closes #306

mb706:
broken because not checked out from master

mb706:
(superseded by #457)

PR #396, opened 2020-04-06
Turn greplicate into a `ppl`

### LINKED ISSUE #385: Turn greplicate into a `ppl`
As this goes with quite a bit of documentation in the book, we might wanna do this at a later point.

sumny:
Warns in `greplicate` that it is deprecated and will be removed soon.
Added pipeline_greplicate and sugar ppl("greplicate"). Moved tests and updated docu. Closes #385


sumny:
book PR: mlr-org/mlr3book#168

PR #405, opened 2020-04-16
PipeOp imputeoor

### LINKED ISSUE #391: out of range imputer
should be named "PipeOpImputeOutOfRange", id "imputeoor". In mlr this was `imputeMin()` or `imputeMax()`. The imputed values should be something like
```r
max(x) + offset + multiplicator * diff(range(x))
```
with `max()` vs `min()`, `offset` and `multiplicator` a parameter defaulting to 1 both. The documentation should stress that this makes sense for tree-based methods and reference [this paper](http://www.jmlr.org/papers/volume11/ding10a/ding10a.pdf).

sumny:
closes #391


berndbischl:
i think i am still standing by my comment in the issue #391 that OOR and NewLevel should be the same pipeop

sumny:
I combined this now with `PipeOpImputeNewlvl`

Issue #402, opened 2020-04-08
autoconvert selector_xxx to po("select")


mb706:

mb706:
Closed by 406

Issue #395, opened 2020-04-05
maybe default po("missind") to only add indicator columns for numerics


mb706:
we usually impute factors with newlvl, so missinds are often not needed.

this would amount to setting `ps$values$affect_columns` in the constructor; would probably have to happen after the `super()` call in the constructor.

Opinions here? @pfistfl  @berndbischl

pfistfl:
agree, taken up in #406

mb706:
Closed by #406

Issue #330, opened 2020-02-17
Selector for factorials by cardinality


mb706:
[This selector](https://github.com/mb706/autoxgboost3/blob/41f3523582792fd332be1a2340ff6e24bfd2562d/R/autoxgboost.R#L162-L168) could be generally useful. I suggest the following changes
* Rename to `selector_cardinality_greater_than`
* Select names where there are *more* levels than given: `levlens > levels`
* Parentheses different in the last line: `names(levlens)[levlens > levels]
* Maybe there is a way to do this without cloning, e.g. by sapplying a `length(distinct_values(x))` on all factorial / logical(?) columns

Prepare this selector in `Selector.R`, see first paragraph in https://github.com/mlr-org/mlr3pipelines/issues/326 how to do that. (I am the original author of autoxgboost3 and I'm granting the necessary license rights for that etc.)

pfistfl:
I did not go deeper into avoiding cloning; would assume that the computational cost does not drastically differ. If you disagree, let me know.

mb706:
I guess it's fine, it probably won't be a hotspot

mb706:
closed by  #406

Issue #320, opened 2020-02-12
Use selectors in select_cols 


mb706:
PipeOps inheriting from PipeOpTaskPreproc(Simple) and overloading the `_dt` functions (e.g. predict_dt, transform_dt) can select the columns they operate on. E.g. PipeOpEncode
```r
select_cols = function(task) {
  task$feature_types[get("type") %in% c("factor", "ordered"), get("id")]
}
```
This selects the features by type. However, we have a shortcut for this now:
```r
select_cols = selector_type(c("factor", "ordered"))
```

To Do:

1. Replace a single select_cols with a `selector_type()` and see if tests still pass. If they don't, get back to me. I hope the tests pass, then just continue.
2. Search for all `select_cols` occurrences in mlr3pipelines and replace them with the fitting `selector_...()` functions. Make sure the correct types are used. There are also some occurrences of
    ```r
    select_cols = function(task) task$feature_names
    ```
    replace those with
    ```r
    select_cols = selector_all()
    ```

mb706:
Please wait until https://github.com/mlr-org/mlr3pipelines/pull/324 is merged before starting this.

pfistfl:
`PipeOpTaskPreproc` and friends are solved centrally via `feature_types`, added selectors for `PipeOpImpute`  and friends.

PR #406, opened 2020-04-16
Selector Issues

pfistfl:
Solves the following issues:
- Autoconvert selector `po(selector_xxx())`: https://github.com/mlr-org/mlr3pipelines/issues/402
- Use selector for `$.select_cols` https://github.com/mlr-org/mlr3pipelines/issues/320
- `cardinality selector`Selector for factorials by cardinality https://github.com/mlr-org/mlr3pipelines/issues/330
- `missind` now selects everything except for `factor, character, ordinal` #395


R/PipeOpImputeHist.R
```diff
@@ -60,7 +60,7 @@ PipeOpImputeHist = R6Class("PipeOpImputeHist",
   ),
   private = list(
 
-    .select_cols = function(task) task$feature_types[get("type") %in% c("numeric", "integer"), get("id")],
+    .select_cols = function(task) selector_type(c("numeric", "integer"))(task),
```
mb706, line 63: I *think* you could jsut do `.select_cols = selector_type(c("numeric", "integer"))` here, but I'm not 100% sure. Have you tried that and it failed for some reason?

pfistfl: So I have not fully understood why, but it fails.

What happens is, that the `types` supplied to `slector_type` are lost after construction.
and I get `Object 'types' not found.`

mb706: interesting, probably R6 takes away environments of methods

mb706: https://github.com/r-lib/R6/issues/80

PR #407, opened 2020-04-18
PipeOp nmf

### LINKED ISSUE #343: Missing PipeOp: Non-negative matrix factorization
#254 See https://github.com/renozao/NMF

sumny:
adds non-negative matrix factorization using `NMF::nmf`
closes #343


sumny:
The following may need some discussion:
```
library("mlr3")
task = tsk("iris")
pop = po("nmf", param_set = list(nrun = 2))
pop$train(list(task))[[1]]$data()
```

yields
`Error in path.package(pkg) : none of the packages are loaded`
even if you `requireNamespace("NMF")`, because `NMF::nmf()` calls `NMF:::setupLibPaths()` when evaluating more than one run, see `?NMF::nmf`, `nrun`. This would require either to import `NMF` or the user to load the library manually.

sumny:
apparently others also have this problem: https://github.com/renozao/NMF/issues/114

sumny:
I debugged this more and apparently this will only be a problem if the nruns are supposed to be evaluated in parallel on NMF's side. I will probably come up with a fix that sets parallel to FALSE which is probably sensible anyhow because of mlr3's outer parallelisation.

Issue #393, opened 2020-04-05
handling of IDs in `ppl()`


mb706:
there should probably be aunified way to handle ID changes of pipeops introduced through `ppl()` / `mlr_graphs`.

pfistfl:
Looking through currently existing `ppl`:
-  `robustify` does not have any, user could just make use of #403 if required
- `bagging` relies on `greplicate` below, does not need any
- `greplicate` should make use of #403 (`update_graph_ids`)
- `branching` currently has two different args which also need to behave differently from other use cases.

mb706:
> user could just make use of #403 if required

I guess that is true in all cases. Let's see how well we like #403. I guess it would make the difference between
```r
x %>>% ppl("whatevs", id_prefix = "y") %>>% z
```
and
```r
x %>>% ppl("whatevs")$prefix_ids("y") %>>% z
```

pfistfl:
PR in #409, we might either wanna use it or decide against it.

pfistfl:
`$update_ids()`

PR #409, opened 2020-04-20
Pre- and postfix ppl on construction

pfistfl:
Adds prefix and postfix to `ppl`: `ppl(xxx, prefix = "foo_")` as suggested in #393.


mb706:
Closing this, because we can already do `ppl(xxx)$update_ids(prefix = "bla_")`

PR #410, opened 2020-04-21
pipeops FilterRows and PredictionUnion

### LINKED ISSUE #301: Feature Request: Filtering Rows of a Task
Use Case: 

I want to split my data into two parts for example according to some logic `x <= 0; x > 0`. 
Then I want to use learner a) on part 1 and learner b) on part 2. 
Afterward, I would like to `cbind` the Predictions. 

This would 1) require a new PipeOp that Filters data according to user-defined logic and
2) A new PipeOp that Cbinds Prediction.

Possible Names: `PipeOpFilterRows`, `PipeOpCbinPrediction`?

sumny:
`PipeOpFilterRows` allows to filter the rows of the data of a task according to a `filter` parameter and adds an easy way for removing rows with missing values, closes #301 
`PipeOpPredictionUnion` allows to combine predictions of the same `task_type` and`predict_type` into a new `Prediction`. This maybe needs some discussion as combining predictions is imo not that straightforward.


R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+#' * `na_column` :: `NULL` | `character`\cr
+#'   A character vector that specifies the columns of the data of the input [`Task`][mlr3::Task]
+#'   that should be checked for missing values. If set to `all`, all columns of the data are used. A
+#'   row is removed if at least one missing value is found with respect to the columns specified.
+#'   Default is `NULL`, i.e., no removal of missing values is done.
+#' * `invert` :: `logical(1)`\cr
+#'   Should the filtering rule be set-theoretically inverted? Note that this happens after
+#'   (possible) missing values were removed if `na_column` is specified. Default is `FALSE`.
+#' * `skip_during_predict` :: `logical(1)`\cr
+#'   Should the filtering and missing value removal steps be skipped during prediction? If `TRUE`,
```
pfistfl, line 63: Shouldn't this be true? I guess we do not filter during predict almost ever?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+#'   How the rows of the data of the input [`Task`][mlr3::Task] should be filtered. This can be a
+#'   character vector of length 1 indicating a feature column of logicals in the data of the input
+#'   [`Task`][mlr3::Task] which forms the basis of the filtering, i.e., all rows that are `TRUE`
+#'   with respect to this column are kept in the data of the output [`Task`][mlr3::Task]. Moreover,
+#'   this can be an expression that will result in a logical vector of length `$nrow` of the data of
+#'   the input [`Task`][mlr3::Task] when evaluated withing the environment of the `$data()` of the
+#'   input [`Task`][mlr3::Task]. Finally, this can also be an integerish vector that directly
+#'   specifies the row identifiers of the rows of the data of the input [`Task`][mlr3::Task] that
+#'   should be kept. Default is `NULL`, i.e., no filtering is done.
+#' * `na_column` :: `NULL` | `character`\cr
```
pfistfl, line 54: Default to `character(0)`

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+#'   A character vector that specifies the columns of the data of the input [`Task`][mlr3::Task]
+#'   that should be checked for missing values. If set to `all`, all columns of the data are used. A
```
pfistfl, line 56: can we somehow use `all` that is unlikely to be a colname? `_all_` or something?

pfistfl: What have we done in other `PipeOp`s here?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+#' * `param_vals` :: named `list`\cr
+#'   List of hyperparameter settings, overwriting the hyperparameter settings that would otherwise
+#'   be set during construction. Default `list()`.
+#'
+#' @section Input and Output Channels:
+#' Input and output channels are inherited from [`PipeOpTaskPreproc`].
+#'
+#' The output during training is the input [`Task`][mlr3::Task] with rows kept according to the
+#' filtering (see Parameters) and (possible) rows with missing values removed.
+#'
```
pfistfl, line 27: Maybe document what happens if `filter` AND `na_column` are set?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+      if (skip) {
+        return(task)  # early exit if skipped (if skip_during_predict)
+      }
+
+      row_ids = task$row_ids
+
+      # NA column(s) handling
+      na = self$param_set$values$na_column
+      if (!is.null(na)) {
+        assert_subset(na, choices = c("all", colnames(task$data())))
```
pfistfl, line 118: more out of curiosity: Should we use `task$feature_names` here?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+        if (na == "all") na = colnames(task$data())
+        na_ids = which(rowSums(is.na(task$data(cols = na))) > 0L)
```
pfistfl, line 120: is this faster then `apply(..., ..., anyNA)` ?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,171 @@
...
+
+      row_ids = if (self$param_set$values$invert) {
+        setdiff(row_ids, filter_ids)
+      } else {
+        intersect(row_ids, filter_ids)
+      }
+
+      # only set the state if required (during training)
+      if (set_state) { 
+        self$state$na_ids = na_ids
```
pfistfl, line 154: why do we make a difference between na_ids and other filtered ids?

tests/testthat/test_pipeop_filterrows.R
```diff
@@ -0,0 +1,90 @@
...
+  task = TaskClassif$new("test", backend = dat, target = "Species")
+
+  op = PipeOpFilterRows$new(param_vals = list(na_column = "all", filter = 1:10))
+  train_out1 = op$train(list(task))[[1L]]
+  expect_equal(op$state$row_ids, setdiff(1:10, c(1, 2, 3, 4, 5)))
+  op$param_set$values$invert = TRUE
+  train_out2 = op$train(list(task))[[1L]]
+  expect_equal(op$state$na_ids, c(1, 2, 3, 4, 5, 100))
+  expect_true(!(100 %in% op$state$row_ids))
+})
```
pfistfl, line 90: We should test how `filterrows` works with learnercv. 
I think we'd like a way to only `train` on a subset of data but get predictions for all of them.
Not sure whether this is actually possible / sensible though.

R/PipeOpFilterRows.R
sumny: Makes sense, I changed the default.

R/PipeOpFilterRows.R
sumny: PipeOpMissingIndicators.R uses `"all"` but I do like the suggestion of `"_all_"`.

R/PipeOpFilterRows.R
sumny: done.

R/PipeOpFilterRows.R
sumny: Then we cannot remove missing rows with respect to the `target` variable (or other variables like `groups` etc.). Do we want this?

R/PipeOpFilterRows.R
sumny: This should be faster as long as p is not >> n, which should be in most of the cases?
```
library(bench)

n <- 1000
p <- 1000
dat <- matrix(sample(c(NA, 0:p, NaN), n*p, TRUE), n, p)
# factor 3 for rowsums
mark(which(apply(dat, 1, anyNA)), which(rowSums(is.na(dat)) > 0L), iterations = 100)

n <- 10000
p <- 1000
dat <- matrix(sample(c(NA, 0:p, NaN), n*p, TRUE), n, p)
# factor 3.5 for rowsums
mark(which(apply(dat, 1, anyNA)), which(rowSums(is.na(dat)) > 0L), iterations = 100)

n <- 1000000
p <- 10
dat <- matrix(sample(c(NA, 0:p, NaN), n*p, TRUE), n, p)
# factor 23 for rowsums
mark(which(apply(dat, 1, anyNA)), which(rowSums(is.na(dat)) > 0L), iterations = 100)

n <- 10
p <- 100000
dat <- matrix(sample(c(NA, 0:p, NaN), n*p, TRUE), n, p)
# factor 1.8 for apply
mark(which(apply(dat, 1, anyNA)), which(rowSums(is.na(dat)) > 0L), iterations = 100)
```

R/PipeOpFilterRows.R
sumny: I kinda liked this separation so I can see directly which were `NA` / `NaN` or which were actually filtered when both is done but I am fine with merging them.

tests/testthat/test_pipeop_filterrows.R
sumny: this should already be possible - i added a toy example in the tests

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,172 @@
...
+#'   How the rows of the data of the input [`Task`][mlr3::Task] should be filtered. This can be a
+#'   character vector of length 1 indicating a feature column of logicals in the data of the input
+#'   [`Task`][mlr3::Task] which forms the basis of the filtering, i.e., all rows that are `TRUE`
+#'   with respect to this column are kept in the data of the output [`Task`][mlr3::Task]. Moreover,
+#'   this can be an expression that will result in a logical vector of length `$nrow` of the data of
+#'   the input [`Task`][mlr3::Task] when evaluated withing the environment of the `$data()` of the
+#'   input [`Task`][mlr3::Task]. Finally, this can also be an integerish vector that directly
+#'   specifies the row identifiers of the rows of the data of the input [`Task`][mlr3::Task] that
+#'   should be kept. Default is `NULL`, i.e., no filtering is done.
+#' * `na_column` :: `character`\cr
```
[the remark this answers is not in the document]
mb706: (or whatever else we had before with expressions)

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,172 @@
...
+#'   A character vector that specifies the columns of the data of the input [`Task`][mlr3::Task]
+#'   that should be checked for missing values. If set to `_all_`, all columns of the data are used. A
+#'   row is removed if at least one missing value is found with respect to the columns specified.
+#'   Default is `character(0)`, i.e., no removal of missing values is done.
+#' * `invert` :: `logical(1)`\cr
```
mb706, line 60: also don't need this imho because the expression / formula would cover this

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,172 @@
...
+#'   Should the filtering rule be set-theoretically inverted? Note that this happens after
+#'   (possible) missing values were removed if `na_column` is specified. Default is `FALSE`.
+#' * `skip_during_predict` :: `logical(1)`\cr
```
mb706, line 63: is there any setting where filtering during predict would *not* break everything horribly?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,172 @@
...
+        ParamLgl$new("invert", default = FALSE, tags = c("train", "predict")),
+        ParamLgl$new("skip_during_predict", default = TRUE, tags = "predict"))
+      )
+      ps$values = list(filter = NULL, na_column = character(0L), invert = FALSE, skip_during_predict = TRUE)
+      super$initialize(id, param_set = ps, param_vals = param_vals)
+    }
+  ),
+  private = list(
+    .na_and_filter = function(task, skip, set_state) {
+      if (skip) {
```
mb706, line 110: this is not good design imho, your function should not have an off-switch, instead you should do `if (skip) task else private$.filter(...)`. But as I say above I really doubt we need the "skip" thing at all and should just not do anything during prediction. (But convince me otherwise if you think there are reasons)

mb706: (well this is an in-place op, so it would be even easier: `if (!skip) private$.filter(....) ; task`

R/PipeOpPredictionUnion.R
```diff
@@ -0,0 +1,136 @@
+#' @title PipeOpPredictionUnion
```
mb706, line 1: 1) please do separate pipeops in separate PRs
2) shouldn't the ensembling pipeops handle this case?

R/PipeOpPredictionUnion.R
```diff
@@ -0,0 +1,136 @@
...
+    }
+  ),
+  private = list(
+    .train = function(inputs) {
+      self$state = list()
+      list(NULL)
+    },
+    .predict = function(inputs) {
+      # currently only works for task_type "classif" or "regr"
+      check = all((unlist(map(inputs[-1L], .f = `[[`, "task_type")) == inputs[[1L]]$task_type) &
```
mb706, line 106: when you know the supposed result of the `map()` call, you can use the `map_xxx` functions, which are basically like `vapply()` only that you have to remember more names. In this case `map_chr()`.

mb706: also with `map()` and its cousins you can just do `map_chr(inputs[-1], "task_type")`

mb706:
I'm not sure how `mlr3` currently handles / should handle `Prediction` with missing rows. We may also want to use classifavg / regavg for unioning because they would support multiple predictions of the same ID.

mb706:
Also the need to filter rows during training (e.g. based on NA or some other property that a learner cannot handle) may be very different from this splitting up predictions between learners thing?

R/PipeOpFilterRows.R
```diff
@@ -0,0 +1,172 @@
...
+#' * `na_ids` :: `integer`\cr
+#'   The row identifiers that had missing values during training and therefore were removed. See the
+#'   parameter `na_column`.
+#' * `row_ids` :: `integer`\cr
+#'   The row identifiers that were kept during training according to the parameters `filter`,
+#'   `na_column` and `invert`.
+#'
+#' @section Parameters:
+#' The parameters are the parameters inherited from [`PipeOpTaskPreproc`], as well as:
+#' * `filter` :: `NULL` | `character(1)` | `expression` | `integer`\cr
```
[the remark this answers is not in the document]
sumny: cleaned this up and provide a formula interface now, which is indeed way more simple. will come up with a `pack_formula()` PR soon.

R/PipeOpPredictionUnion.R
sumny: dropped the union stuff of predictions from this PR for now

sumny:
Cleaned this up. Currently has two parameters:
* `filter_formula` (`formula` with only a rhs evaluating to `TRUE` or `FALSE` within the data of the `Task` indexing rows to keep)
* `na_selector` (`Selector` function selecting features that should be checked for missing values results in rows having missing values w.r.t to these features being dropped)

I agree with @mb706 that this is somewhat redundant, because:
```
task = tsk("pima")
po1 = po("filterrows", param_vals = list(filter_formula = ~ !is.na(insulin)))
po2 = po("filterrows", param_vals = list(na_selector = selector_name("insulin")))
```

results in virtually the same operation being done.
However, what if I have a huge number of features and want to e.g. remove missing values w.r.t to `factorial` features? The approach via `filter_formula = ~ !is.na(feature1) & !is.na(feature2) & ...` seems quite cumbersome compared to `na_selector = selector_type("factor")`? I guess you could somewhat auto-generate the formula in this case but I am not sure if this is really nicer..

mb706:
That argument for 'nacols" would also apply to other conditions to filter by, e.g. trying to filter if any of a number of features is 0 or negative. We could go the `data.table` row and
1. Introduce the `.SD` variable that is just a `data.table` of the dataset
1. introduce an `SDcols` hyperparameter that selects the features that should be present in `.SD`

So then dropping rows where any feature is `NA` would go down to
```r
SDcols = selector_grep("^nonNA\\.")
filter_formula = ~ !apply(is.na(.SD), 1, any)
```

mb706:
So this is supposed to filter out rows during predict time, right? Is there a way to keep track of what sample is being predicted, once a learner sees the incomplete prediction dataset? So if my input task has two rows, one is dropped, my Graph returns one prediction, which sample does the prediction belong to?

Suppose that is not a problem, what is the current state of `mlr3` handling missing predictions? ignores them? uses a fallback learner? just crashes? In the latter case we whould modify `GraphLearner` to do something sensible with missing predictions.

Also, maybe this `PipeOp` have the option to select on what phase samples are filtered? E.g. `"train"`, `"predict"`, `"always"`.

sumny:
> That argument for 'nacols" would also apply to other conditions to filter by, e.g. trying to filter if any of a number of features is 0 or negative. We could go the `data.table` row and
> 
>     1. Introduce the `.SD` variable that is just a `data.table` of the dataset
> 
>     2. introduce an `SDcols` hyperparameter that selects the features that should be present in `.SD`
> 
> 
> So then dropping rows where any feature is `NA` would go down to
> 
> ```r
> SDcols = selector_grep("^nonNA\\.")
> filter_formula = ~ !apply(is.na(.SD), 1, any)
> ```

I like the idea of using `SDcols`. I included this. So filtering is now only specified using the `filter_formula` which can rely on the `.SD` variable because it is evaluated within the frame of the `data.table` backend of the `Task`.

sumny:
> So this is supposed to filter out rows during predict time, right? Is there a way to keep track of what sample is being predicted, once a learner sees the incomplete prediction dataset? So if my input task has two rows, one is dropped, my Graph returns one prediction, which sample does the prediction belong to?
> 
> Suppose that is not a problem, what is the current state of `mlr3` handling missing predictions? ignores them? uses a fallback learner? just crashes? In the latter case we whould modify `GraphLearner` to do something sensible with missing predictions.
> 
> Also, maybe this `PipeOp` have the option to select on what phase samples are filtered? E.g. `"train"`, `"predict"`, `"always"`.

I think we are probably mixing up different use cases. I believe what @pfistfl wanted to be possible in #301 is to simple split your task based on some logic both during training and prediction (with the row ids of the splits being disjunct and their union being equal to the row ids of the complete task). In this case using `PipeOpFilterRows` would look somewhat like this:

```
task = tsk("pima")
gr1 = po("filterrows", param_vals = list(filter_formula = ~ age < 31)) %>>% lrn("classif.rpart")
gr2 = po("filterrows", param_vals = list(filter_formula = ~ age >= 31)) %>>% lrn("classif.rpart")
gr1$train(task)
gr2$train(task)
p1 = gr1$predict(task)[[1L]]  # correct row_ids are pertained 
p2 = gr2$predict(task)[[1L]]
setequal(union(p1$row_ids, p2$row_ids), task$row_ids)
```

Then we would only need another `PipeOp` that would do nothing during training and combine the predictions during prediction, i.e., simply wrapping `c` so we could have something like `copy %>>% list(filterrows1, filterrows2) %>>% unite` and the prediction output would be:

```
c(p1, p2)
```

```
<PredictionClassif> for 768 observations:
    row_id truth response
         4   neg      neg
         6   neg      neg
         7   pos      neg
---                      
       763   neg      neg
       764   neg      neg
       767   pos      pos
```

py9mrg:
Hello,

I thought this new feature might help my issue #566. I gave it a try using this commit and then adding the line `po("filterrows", param_vals = list(filter_formula = ~ !apply(is.na(.SD), 1, any)))` before my svm learner. But I still got the warning about mismatching truth and response lengths so either I'm doing something wrong or this won't solve #566. Please let me know if I'm being really thick!

sumny:
Hi @py9mrg,

your approach above does not work out of the box because the `SDcols` hyperparameter of `PipeOpFilterRows` is set to `selector_all()` by default, and `selector_all()` called on a `Task` will only return the feature names of the task. As you deal with missing values in your target variable you can provide your own selector, e.g.:

```r
selector_all_inc_target = function() {
  make_selector(function(task) {
    c(task$feature_names, task$target_names)
  }, description = "selector_all_inc_target")
}
```

then use this in your pipeop:

```r
po = po("filterrows",
  param_vals = list(
    filter_formula = ~ !apply(is.na(.SD), 1, any),
    SDcols = selector_all_inc_target()
  )
)
```

this would result in the 3rd row also being dropped:

```r
po$train(list(task_w_missing))[[1]]$data()
```

```
    target variable1 variable2
 1:     32         4         4
 2:     50         5         5
 3:     72         6         6
 4:     98         7         7
...
```

All in all this should work for you.
However, there is some inconsistency now that the state of the `po` would only list your features `"variable1"` and `"variable2"` as being affected (although your target is also). We may have to discuss what to do with `PipeOp`s affecting targets in general. Also I will try to finish this PR up soon and maybe refactor it. Hope this helps.

py9mrg:
Hello @sumny ,

Thanks so much for coming back to me. That works! Just one thing in case anyone is reading with a similar use case, because `make_selector` is not exported, I had to use `mlr3pipelines:::make_selector`. Took me longer than I care to admit to work that out.

Thanks again for your help.

py9mrg:
Hello @sumny, do you know when this PR might be merged? I'm using it all the time and it would be nice, when not using renv, to not have to keep rolling mlr3pipelines back to this commit every time I forget to untick the update box in RStudio for this particular package!

sumny:
@py9mrg, @mb706 and I have to go over it and check whether the complete functionality is still within the scope of mlr3pipelines. Also, selecting targets via selectors is still an open discussion #493 I will ping you once it is merged - sorry for keeping you waiting so long!

py9mrg:
@sumny thanks and don't worry - I'm sure I am completely underestimating how significant the change is.

PR #412, opened 2020-04-23
add neutral elements to %>>%

### LINKED ISSUE #328: Should %>>% have a neutral element?
I have seen enough instances of `if (!is.null(another_po)) my_pipeline = my_pipeline %>>% another_po` that I think it would be nice to have a placeholder for `%>>%` that makes it not do anything. mlrCPO has the [NULLCPO](https://github.com/mlr-org/mlrCPO/blob/master/R/NULLCPO.R) which I really like but for some reason I always had a hard time convincing some people it is useful. I am therefore putting this up for discussion, do you think this should exist?

An example usecase would be [this](https://github.com/mlr-org/mlr3pipelines/pull/323/files#diff-ef9e8b3b60c4d22015856274ef37b03aR159-R160).

sumny:
e.g.:
`p1 %>>% p2`
if `p1` is `NULL` return `as_graph(p2)`
if `p2` is `NULL` return `as_graph(p1)`

closes #328


mb706:
thsi should be documented then it is fine

sumny:
> thsi should be documented then it is fine

done!

PR #413, opened 2020-04-23
PipeOpTargetTrafoSimple tests fix

sumny:
Currently `master` is broken due to the test of `PipeOpTargetTrafoSimple` where we only change the target name but keep the `identity` as the trafo/inverter function. This conflicts with the early exit handling of `PipeOpTargetTrafoSimple` and imo it is not sensible to use `PipeOpTargetTrafoSimple` if you only want to set new target names (you should use `PipeOpUpdateTarget` in this case). Therefore, I removed this test.


sumny:
This now further adds an early exit to `PipeOpUpdateTarget` to allow for simply setting an existing column in the data as the new target. Needed for mlr-org/mlr3gallery#27

PR #415, opened 2020-04-27
added flag for (not) dropping levels while prediction

Coorsaa:


Coorsaa:
merge of branch `ytrafo` makes PR #413 obsolete

PR #416, opened 2020-04-28
Pipeop imputeconstant

### LINKED ISSUE #392: PipeOpImputeConstant
this one is *different* from imputenewlvl and may in fact not introduce a new level for missing values. The parameter should be a list of arbitrary length with formula elements. The usage should be
```r
pop = po("imputeconst")
pop$param_set$values$imputing = list(
  selector_name("Sepal.Length") ~ 1,
  selector_grep("^\\.Width") ~ -1,
  selector_type("factor") ~ "x"
)
```
note we are working with `formula` here, and we can impute different values for different columns defined by their selectors. (earlier selectors overwrite later selectors here).

Note that there are `Selector`s that are data-dependent. Therefore the "model" should be to collect the actual column names that get imputed in a certain way, depending on the selectors. We may need to overload `PipeOpImpute` in an unusual way, write me if this gets difficult.

sumny:
closes #392 
design follows the latest [comment](https://github.com/mlr-org/mlr3pipelines/issues/392#issuecomment-616078541).


R/PipeOpImputeConstant.R
```diff
@@ -0,0 +1,112 @@
...
+#' @family Imputation PipeOps
+#' @include PipeOpImpute.R
+#' @export
+PipeOpImputeConstant = R6Class("PipeOpImputeConstant",
+  inherit = PipeOpImpute,
+  public = list(
+    initialize = function(id = "imputeconstant", param_vals = list()) {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("constant", tags = c("train", "predict", "required"), custom_check = function(x) {
+          check_atomic_vector(x, any.missing = FALSE, len = 1L)
```
mb706, line 71: checkScalar

R/PipeOpImputeConstant.R
```diff
@@ -0,0 +1,112 @@
...
+
+    .impute = function(feature, type, model, context) {
+      constant = self$param_set$values$constant
+      switch(type,
+        "logical"   = assert_flag(constant),
+        "integer"   = assert_int(constant),
+        "numeric"   = assert_number(constant),
+        "character" = assert_string(constant),
+        "factor"    = assert_string(constant),
+        "ordered"   = assert_string(constant),
```
mb706, line 94: these should probably all also allow factors

R/PipeOpImputeConstant.R
```diff
@@ -0,0 +1,112 @@
...
+        "POSIXct"   = assert_posixct(constant, any.missing = FALSE, len = 1L)
+      )
+
+      if (type == "integer") {
+        constant = as.integer(constant)
+      }
+
+      if (type %in% c("ordered", "factor")) {
+        levels(feature) = c(levels(feature), constant)
```
mb706, line 103: maybe add a flag if new levels are something the user wants. if it is false (default) then there is an error if the level is not present in ordered / factor features. (ignore this for characters)

Issue #414, opened 2020-04-26
check if pipeophistbin works correctly if not all bins are present


mb706:

pfistfl:
I guess this would be solved by writing a unit test for this.

ZackBarry:
I'd like to add a PR for this.  Just to clarify - is this the sequence of events that should be tested? 

1) A PipeOpHistBin is trained on Task1 , which has a value in (a,b]
2) The output from (1) does not contain (a,b]
3) The trained po is used for prediction on Task2, which has a value in (a,b]
4) The output from (3) contains (a,b]

mb706:
Closed by https://github.com/mlr-org/mlr3pipelines/pull/372 (predict time value out of range) and #421 (predict time value in train-time-empty-bin)

PR #421, opened 2020-05-05
test for case where not all bins present in training data

ZackBarry:
This test is in reference to [https://github.com/mlr-org/mlr3pipelines/issues/414](https://github.com/mlr-org/mlr3pipelines/issues/414). Test this sequence of events works correctly:

1. A PipeOpHistBin is trained on Task1 , which has a value in (a,b]
2. The output from (1) does not contain (a,b]
3. The trained po is used for prediction on Task2, which has a value in (a,b]
4. The output from (3) contains (a,b]

### DIFF
--- a/tests/testthat/test_pipeop_histbin.R
+++ b/tests/testthat/test_pipeop_histbin.R
@@ -46,3 +46,25 @@ test_that("PipeOpHistBin - numerics out of range of training data", {
     as.logical(grep("-Inf", x[1])) && as.logical(grep("Inf", x[length(x)]))
   })))
 })
+
+test_that("PipeOpHistBin - not all bins present", {
+  task1 = mlr_tasks$get("iris")
+  dat = iris
+  dat$Sepal.Width[[1L]] = 2.13
+  task2 = TaskClassif$new("iris2", backend = dat, target = "Species")
+  
+  op = PipeOpHistBin$new(param_vals = list(breaks = seq(0, 10, by = 0.05)))
+  expect_pipeop(op)
+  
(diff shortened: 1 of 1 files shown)


ZackBarry:
The check is failing on a separate test, but it doesn't seem as if this is due to my changes.  Here's the line the check failure references: [https://github.com/mlr-org/mlr3pipelines/blob/98640e69a7acc4f5bfa0b4037834b096ce16c2e5/tests/testthat/test_pipeop_targettrafosimple.R#L41](https://github.com/mlr-org/mlr3pipelines/blob/98640e69a7acc4f5bfa0b4037834b096ce16c2e5/tests/testthat/test_pipeop_targettrafosimple.R#L41).

sumny:
@ZackBarry thanks a lot for your first PR and your interest in mlr3pipelines! We currently have an internal workshop and look into this. It may be that we already fixed parts of this in #372 but we will come back to you. Thanks again!

sumny:
> The check is failing on a separate test, but it doesn't seem as if this is due to my changes. Here's the line the check failure references:
> 
> https://github.com/mlr-org/mlr3pipelines/blob/98640e69a7acc4f5bfa0b4037834b096ce16c2e5/tests/testthat/test_pipeop_targettrafosimple.R#L41
> .

this was indeed not due to your changes but a bug from our side - maybe you can merge the current master into your PR and see if the CI passes.

ZackBarry:
@sumny Glad to see the checks are working! I'm not sure how to read the Codecov Report, but I was suprised to see the coverage go down. Am I missing something?

mb706:
Thanks, took me a while to get that this tests for bins not present *within* the training-range of a feature.

PR #422, opened 2020-05-06
Fix call to po() erroring

mllg:
A call to `po()` is currently erroring (in contrast to `lrn()` or other sugar functions).


mb706:
does this do essentially the same as https://github.com/mlr-org/mlr3pipelines/pull/434 ? should I merge this one and revert the other one?

mllg:
Sorry, I forgot I already did a PR for this. Both approaches are fine AFAICT.

Issue #420, opened 2020-05-05
Renaming features: task$rename_features (moved from mlr3)


berndbischl:
Posted by @ZackBarry here (and moved by me):
https://github.com/mlr-org/mlr3/issues/268#issuecomment-623784752

----

Is it possible to use the task$rename() method within a pipeline? One use case that I haven't been able to find any solutions for is to union the results of two PCA branches. When I've tried this with gunion and mlr_pipeops_featureunion, only one set of PCX columns remains at the end of the graph. I'd like to be able to rename the output columns from the left PCA branch as "left_PCX" and the right branch as "right_PCX".

Toy example:
`library(mlr3)
library(mlr3pipelines)
library(dplyr)

toy_df <- data.frame(
homeWins = c(1, 0, 1, 0),
homeHits = c(13,10,12,11),
homeHomeRuns = c(0,2,0,0),
awayHits = c(10,11,14,8),
awayHomeRuns = c(1,0,1,1)
) %>%
mutate(homeWins = as.factor(homeWins))

task = TaskClassif$new("toy_task", backend = toy_df, target = "homeWins", positive = "1")

pca_home = po("select", id = "home_cols", param_vals = list(selector = selector_grep("home"))) %>>%
po("pca", id = "home_pca", param_vals = list(center = TRUE, scale. = TRUE))

pca_away = po("select", id = "away_cols", param_vals = list(selector = selector_grep("away"))) %>>%
po("pca", id = "away_pca", param_vals = list(center = TRUE, scale. = TRUE))

graph <- gunion(list(pca_home, pca_away)) %>>%
po("featureunion", id = "all_together")

graph$keep_results <- TRUE

graph$train(task)
#> $all_together.output
#> TaskClassif:toy_task (4 x 3)
#> * Target: homeWins
#> * Properties: twoclass
#> * Features (2):
#> - dbl (2): PC1, PC2`

berndbischl:
@mb706 
Ok, I looked into this.

I think currently this is not (easily) possible, although you could EASILY create a local PipeOp that does this now. But we should support this out-of-the box to ensure flexible usage of the package

Options:

1) Allow a PipeOpRename type of op. 

2) Add a col-renaming function to the PipeOpPreproc-Task base class. 
(that would potentially rename cols after the preproc abstract method is called)

3) Let PipeOpColApply also handle that, maybe it already can. At least that's not documented.

[...]

PR #429, opened 2020-05-27
add PipeOpRenameColumns

sumny:
see also #420


mb706:
one parameter: named character `renaming`. renames columns as found in *names* of `renaming`  to *values* of `renaming`.

mb706:
maybe an extra flag, if we allow columns in the `renaming` vector that are not actually present in the task.

sumny:
> one parameter: named character `renaming`. renames columns as found in _names_ of `renaming` to _values_ of `renaming`.

done!

sumny:
> maybe an extra flag, if we allow columns in the `renaming` vector that are not actually present in the task.

This will nevertheless be checked when the `rename` mutator of the `Task` is being called, so I am not sure whether we need this.

PR #433, opened 2020-06-08
Support lgr

mllg:
Adds initial support for logging via lgr.


R/zzz.R
```diff
@@ -20,6 +20,11 @@ register_mlr3 = function() {
   register_mlr3()
   setHook(packageEvent("mlr3", "onLoad"), function(...) register_mlr3(), action = "append")
   backports::import(pkgname)
+
+  assign("lg", lgr::get_logger("mlr3/mlr3pipelines"), envir = parent.env(environment()))
+  if (Sys.getenv("IN_PKGDOWN") == "true") {
```
mb706, line 25: this is a bit unsustainable; arent the examples also built in other places where we want clean output? If we don't want the info outputs in our demo code, why do we assume that anyone would want them? Why don't we put the default to "warn"?

mllg: Well, we usually want the output. However, our current CI setup does not push to an orphan branch (*sigh*). If we print the logs during pkgdown, we unnecessarily blow up the history due to the timestamps.

PR #439, opened 2020-06-19
minor doc changes to targettrafos, rerun docs

sumny:
some minor changes to the docs of targettrafo pipeops.

Also some more thoughts after writing mlr-org/mlr3gallery#57

*   `train_target()` in `PipeOpTargetTrafo` should probably be renamed to something like `transformer()` (making more clear what should happen here)

*   `PipeOpTargetTrafo` should make use/allow for overloading `task_type` so that subclasses can restrict themselves to different task types

*   to update a task with respect to the target we use `convert_task` (which we should sometime move to `mlr3` and test better); to update a prediction, users need to manually construct a new `Prediction` object because truth and response are read only. Maybe we should provide a higher level function like `convert_prediction` that allows for this more easier

* wrapping a `Graph` consisting of a `PipeOpTargetTrafo %>>% Learner %>>% PipeOpTargetInverter` in a `GraphLearner` requires to manually set the `task_type` during construction of the `GraphLearner` because guessing the type is not possible when `PipeOpTargetInverter` is last in the chain.

### DIFF
--- a/R/PipeOpTrafo.R
+++ b/R/PipeOpTrafo.R
@@ -17,18 +17,20 @@
 #'   `train_target()`.
 #'
 #' * `train_target()` has a [`Task`][mlr3::Task] input and should return a modified
-#'   [`Task`][mlr3::Task]. This typically consists of calculating a new target and modifying the
-#'   task by using `private$.update_target()`. `train_target()` will be called during training and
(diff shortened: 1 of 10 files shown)


sumny:
superseded by mlr-org/mlr3pipelines#440

PR #440, opened 2020-06-21
Straighten out some issues with ytrafo & co

mb706:
* names that are consistent with PipeOpTaskPreprocSimple
* functions in `private$` instead of `self$`
* reasonable default behaviour for some functions
* `.invert()` does the actual inversion instead of returning a function (?!)
* obviate some hyperparameters that instead are instead inferred from table column names
* PipeOpUpdateTarget should probably go in the attic?


sumny:
I merged some older doc changes into this and fixed all tests etc. Some notes:

* in `PipeOpTargetMutate` the reworked design simply passes the original `$truth()` of the input task through (via `predict_phase_state`) so during the target inversion we can rely on this to simply set the `truth` of the new `Prediction`. I like this because this does not involve transforming and inverting the `truth` which could be numerically unstable. However, this does no longer allow for things such as done in the tests here: (I guess this was a nice to have "gimmick" but If somebody really wants to do this, he/she should writes his/her own subclass of `PipeOpTargetTrafo`?)

*   to update a task with respect to the target we use `convert_task` (which we should sometime move to `mlr3` and test better); to update a prediction, users need to manually construct a new `Prediction` object because `$truth` and `$response` are read only. Maybe we should provide a higher level function like `convert_prediction` that allows for this more easier? Or are we fine with the current design?

* wrapping a `Graph` consisting of a `PipeOpTargetTrafo %>>% Learner %>>% PipeOpTargetInvert` in a `GraphLearner` requires to manually set the `task_type` during construction of the `GraphLearner` because guessing the type is not possible when `PipeOpTargetInvert` is last in the chain. I guess this is fine, I just wanted to highlight this.

R/PipeOpTrafo.R
```diff
@@ -45,168 +27,200 @@
...
+      # PipeOpTargetInvert will later apply this function during prediction
       function(inputs) {
         assert_list(inputs, len = 1L, types = "Prediction")
-        self$inverter(inputs[[1L]], predict_phase_control)
+        list(private$.invert(inputs[[1L]], predict_phase_state))
       }
+    },
+
+    # updates the target of a task and also the task_type (if needed), uses convert_task
+    .update_target = function(task, new_target, new_type = NULL, ...) {
```
mb706, line 181: When I thought up the ".update_target" function I thought it would contain all the business code, but that is now in convert_task. I think a better solution now would be to just have users call `convert_task` inside their `.transform()` function, and to remove the `private$.update_target` thing altogether.

R/PipeOpTrafo.R
```diff
@@ -277,111 +290,108 @@ mlr_pipeops$add("targetinverter", PipeOpTargetInverter)
...
+      ))
+      # We could add a condition here for new_task_type on trafo and inverter when mlr-org/paradox#278 has an answer.
+      # HOWEVER conditions are broken in paradox, it is a terrible idea to use them in PipeOps,
+      # see https://github.com/mlr-org/paradox/issues/216 and related comment in PipeOpLearnerCV
 
-    train_target = function(task) {
-      # handle trafo = identity separately
-      if (identical(self$param_set$values$trafo, identity)) {
-        return(task)  # early exit
+      ps$values = list(trafo = identity, inverter = function(x) list(response = identity(x$response)))
```
mb706, line 364: why does `identity` not work? because `x` still contains the `$truth` column? In that case `invert` should probably remove the `truth` col and make `identity` work here.

mb706:
> Maybe we should provide a higher level function like convert_prediction that allows for this more easier?

I don't think that is so useful because the prediction is basically just a list of `$truth`, `$prob` and `$response` with format checking. The only benefit I see is tt would make the

> `get(mlr_reflections$task_types[type]$prediction)$new`

marginally shorter in case we don't know the type of the prediction yet -- which is not enough to justify a new function imho.

mb706:
> guessing the type is not possible when PipeOpTargetInvert is last in the chain

image

I think this is a necessary functionality that should be similar to https://github.com/mlr-org/mlr3pipelines/pull/394, and in fact I think https://github.com/mlr-org/mlr3pipelines/pull/394 should be extended to allow making all properties of POs somehow a function of predecessors in the graph. Then the Inverter could query the predecessor along the `"fun"` channel and would

R/PipeOpTrafo.R
sumny: Makes sense, I will fix this!

R/PipeOpTrafo.R
sumny: because `x` also contains the `row_id` as a column. So I guess I will simply drop this as well

sumny:
> > However, this does no longer allow for things such as done in the tests here
> 
> what do you mean here?

I forgot to hyperlink.. [here](https://github.com/mlr-org/mlr3pipelines/blob/9f7298bc03e2add3ba5aed666169683a790fa7eb/tests/testthat/test_pipeop_targetmutate.R#L70) and [here](https://github.com/mlr-org/mlr3pipelines/blob/9f7298bc03e2add3ba5aed666169683a790fa7eb/tests/testthat/test_pipeop_targetmutate.R#L96) Essentially transformations where the truth is changed in a way that we also must transform.

sumny:
closes #447

sumny:
I added functionality in `GraphLearner` to recursively infer the `task_type` if it fails to do this from the `input` and `output` of the `Graph` (somewhat similary to #394). I guess we could discuss whether adding an active binding `task_type` makes sense but this may be over the top and inferring the `task_type` also for target transformation `Graph`s  mostly works now.

mb706:
merging this and fixing the error together with #443

PR #442, opened 2020-06-21
learner ABs

### LINKED ISSUE #297: Learner should be saved with it's model
Not entirely sure how we want to solve this, but things currently are
 a bit inconsistent between mlr3 and mlr3pipelines.

**Use-Case:**
My learner defines a `plot` method that can be used after training.
I would basically use `learner$train(task); learner$plot()` to call the plot after training. 
This works fine for standard mlr3 learners.
Internally the plot function would do something like 
```
plot = function() plot(self$model)
```

If I now use this with pipelines, the `model` is not saved inside the learner, but inside the pipeop's state.
This would mean that all methods defined inside the learner do no longer work?

**Example**

```r
library(mlr3pipelines)
polrn = PipeOpLearner$new(lrn("classif.rpart"))
polrn$train(list(mlr_tasks$get("iris")))
polrn$state$model
polrn$learner$model
```

>  polrn$state$model
> n= 150 node), split, n, loss, yval, (yprob)
>      * denotes terminal node
> ...

> polrn$learner$model
NULL

mb706:
closes #297


mb706:
@sumny could you write tests for this?

sumny:
I added some tests but the following needs some thoughts:

```
task = tsk("iris")
learner = po("learner", lrn("classif.rpart"))
learner$learner$predict_type = "prob"
learner$train(list(task))
learner$predict(list(task))

learner$learner$predict_type = "response"
```

```
Error in (function (val)  : $learner is read-only.
```

i.e., once we have trained the pipeop/set the state the first time, we can no longer access the *real* learner stored in `private$.learner` and set its `predict_type`, because we return a [deep cloned](https://github.com/mlr-org/mlr3pipelines/blob/ac88d657bdf97418f149a4d2bc5aadaf555cfd1e/R/PipeOpLearner.R#L111) copy of it.

mb706:
I guess we would have to solve this with #394 and let the `PipeOpLearner` have its own `predict_type`

mb706:
or we could have a `$learner` and a `$trained_learner` slot?

mb706:
(where `$learner` is just the empty learner and `$trained_learner` is the AB with the model)

mb706:
For the record, there was the suggestion by @pfistfl to have the `$learner` be the location where the Learner `$state` is stored. This is how `Graph` works with respect to pipeops. Trouble is that the `$state` also gets set to non-standard values through things like branching (set to `NO_OP`) and multiplicities (set to a `Multiplicity`). I guess the best solution is to have a `$learner`, which always gives the *empty* learner that can be manipulated, and a `$learner_model`, that contains the learner + model, but may also contain `NULL` if `NO_OP` was received, or multiple `Learners` if a multiplicity was evaluated. This way one can access the different `Learner`s that were evaluated in multiplicities.

sumny:
So `$learner` now behaves like it used to. And there is now also `$learner_model` that contains the learner + model if trained.

PR #443, opened 2020-06-22
Multiplicities (virtual greplicate)

### LINKED ISSUE #293: virtual greplicate
Maybe it would be possible without much pain to get bagging to work from the Graph itself by introducing some kind of do-the-following-multiple-times PipeOp, and another PipeOp that then collects the multiple outputs; this would require some adjustments to the Graph similar to PipeOp(Un)Branch.

mb706:
closes #293
## See #293 for what this is about

* Multiplicity handling
* new POs:
    * PipeOpMultiplicityExply, PipeOpMultiplicityImply
    * PipeOpReplicate
    * PipeOpOVRSplit, PipeOpOVRUnite
* PipeOpEnsemble (PipeOpClassifAvg, PipeOpRegrAvg), PipeOpFeatureUnion: get `collect` argument

Try it:
```r
gr = po("replicate", reps = 10) %>>% po("subsample") %>>%
  lrn("classif.rpart") %>>% po("classifavg", collect = TRUE)
gr$train(tsk("iris"))
#> $classifavg.output
#> NULL
#> 
gr$predict(tsk("iris"))
#> $classifavg.output
#> <PredictionClassif> for 150 observations:
#>     row_id     truth  response
#>          1    setosa    setosa
#>          2    setosa    setosa
#>          3    setosa    setosa
#> ---                           
#>        148 virginica virginica
#>        149 virginica virginica
#>        150 virginica virginica
#> 
gr$state$classif.rpart[1:2]
#> [[1]]
#> [[1]]$model
#> [...]
#> 
#> [[2]]
#> [[2]]$model
#> [...]
#> 
```
#### One vs. Rest classification
```r
gr <- po("ovrsplit") %>>% lrn("classif.rpart") %>>% po("ovrunite")
```
#### OVR with bagged Learners
nested multiplicities!
```r
gr <- po("ovrsplit") %>>%
  po("replicate", reps = 3) %>>%
  po("subsample") %>>% lrn("classif.rpart") %>>%
  po("classifavg", collect = TRUE) %>>%
  po("ovrunite")
```

#### Bagging outside OVR
nested again, but the other way around
```r
gr <-   po("replicate", reps = 3) %>>%
  po("ovrsplit") %>>%
  po("subsample") %>>% lrn("classif.rpart") %>>%
  po("ovrunite") %>>%
  po("classifavg", collect = TRUE)
```

#### Mixing multiplicities with non-multiplicities
Stacking linear model and ranger. The stochastic ranger is fitted multiple times (reps = 3), the deterministic linear model only once. `featureunion` has one multiplicity input and two non-multiplicity inputs, so the non-multiplicity inputs are "recycled".
```r
gr <- list(
    po("nop"),
    po("learner_cv", lrn("regr.lm")),
    po("replicate", reps = 3) %>>% po("learner_cv", lrn("regr.ranger"))
  ) %>>%
  po("featureunion") %>>% lrn("regr.rpart") %>>% po("regravg", collect = TRUE)
```
image
#### World's slowest Random Forest
(is not actually a random forest, because each tree always splits on the *same* `mtry` variables)
```r
ntree <- 10
mtry <- 2

gr <- po("replicate", reps = ntree) %>>%
  po("select", selector = function(task) sample(task$feature_names, mtry)) %>>%
  po("subsample") %>>% lrn("classif.rpart") %>>%
  po("classifavg", collect = TRUE)
```


mb706:
@mllg maybe you also want to have a look at this.

R/PipeOp.R
```diff
@@ -239,6 +239,10 @@ PipeOp = R6Class("PipeOp",
         self$state = NO_OP
         return(named_list(self$output$name, NO_OP))
       }
+      unpacked = unpack_multiplicities(input, multiplicity_type_nesting_level(self$input$train), self$input$name, self$id)
+      if (!is.null(unpacked)) {
+        return(evaluate_multiplicities(self, unpacked, "train", NULL))
```
sumny, line 244: if we always train with an `instate` of `NULL`, then a previously already trained `PipeOp` may be completely reset to a state of `NULL`, if evaluating multiplicities fails during training, see https://github.com/mlr-org/mlr3pipelines/blob/97f38fa14ece48ae887d7bd8e7e1a1d4d7b88386/tests/testthat/test_multiplicities.R#L115

R/PipeOpClassifAvg.R
```diff
@@ -93,15 +85,16 @@ PipeOpClassifAvg = R6Class("PipeOpClassifAvg",
...
+        prob = weighted_factor_mean(map(inputs, "response"), weights, lvls)
+      } else {
+        stop("PipeOpClassifAvg input predictions had missing 'prob' and missing 'response' values. At least one of them must be given for all predictions.")
       }
 
-      if (has_classif_response) {
-        response = weighted_factor_mean(map(inputs, "response"), weights, lvls)
-      }
-
-      PredictionClassif$new(row_ids = row_ids, truth = truth, response = response, prob = prob)
```
sumny, line 104: I guess it is on purpose to now always return `prob` and calculate `response` based on this, i.e., the following test should now be fixed? https://github.com/mlr-org/mlr3pipelines/blob/97f38fa14ece48ae887d7bd8e7e1a1d4d7b88386/tests/testthat/test_pipeop_ensemble.R#L89

R/PipeOpEnsemble.R
```diff
@@ -151,10 +166,11 @@ weighted_matrix_sum = function(matrices, weights) {
 # @return `factor`
 weighted_factor_mean = function(factors, weights, alllevels) {
```
sumny, line 167: this used to return a `factor` with the levels as arg max with respect to the `prob` matrix and is used this way by `LearnerClassifAvg`: https://github.com/mlr-org/mlr3pipelines/blob/97f38fa14ece48ae887d7bd8e7e1a1d4d7b88386/tests/testthat/test_learner_weightedaverage.R#L12 Maybe we should add a flag to have a fallback to this return value? Or rather change `LearnerClassifAvg` directly?

R/PipeOpClassifAvg.R
```diff
@@ -71,20 +71,12 @@
 PipeOpClassifAvg = R6Class("PipeOpClassifAvg",
   inherit = PipeOpEnsemble,
   public = list(
-    initialize = function(innum = 0, id = "classifavg", param_vals = list()) {
-      super$initialize(innum, id, param_vals = param_vals, prediction_type = "PredictionClassif", packages = "stats")
+    initialize = function(innum = 0, collect = FALSE, id = "classifavg", param_vals = list()) {
```
sumny, line 74: is/should there be any difference in the return value when used with `collect = FALSE`? Just out of curiosity. I guess not, so this simply allows to work for `Multiplicities`?

R/PipeOpClassifAvg.R
mb706: yes, I think this was written when response / prob still worked slightly different in mlr3

R/PipeOpEnsemble.R
mb706: I think LearnerClassifAvg should be changed

R/PipeOpClassifAvg.R
mb706: The "collect" only makes the PO "multiplicity-aware", so that it treats multiplicities explicitly. It does not change the return object (other than the fact that multiplicity-oblivious POs return multiplicities when given multiplicities etc., but that is a boring detail)

R/PipeOp.R
mb706: What happens with the state of a pipeop when training fails is undefined currently. We could agree on "on failure we reset to `NULL`, because the pipeop is effectively not trained then, but that should probably be implemented and tested separately. Here the `NULL` behaviour is incidental so it probably shouldn't be tested in this PR. you could open an issue (and possibly a PR) about this.

mb706:
merging this and fixing the error together with #440

PR #451, opened 2020-07-23
Adds phase hyper-parameter to PipeOpNOP

RaphaelS1:
The purpose of this is to allow control over which phases information is passed through NOP. If `train` or `both` then training returns the input, otherwise NULL. Analogously for `predict`.

### DIFF
--- a/R/PipeOpNOP.R
+++ b/R/PipeOpNOP.R
@@ -26,7 +26,10 @@
 #' The `$state` is left empty (`list()`).
 #'
 #' @section Parameters:
-#' [`PipeOpNOP`] has no parameters.
+#' * `phase` :: `character(1)`\cr
+#' What phase information should be pushed through the `PipeOp`. If `"train"` then in training the
+#' output is identical to input but in prediction the output is `NULL`. Analogously for `"predict"`.
+#' Default is `"both"` so objects are pushed through both training and predicting.
 #'
 #' @section Internals:
 #' [`PipeOpNOP`] is a useful "default" stand-in for a [`PipeOp`]/[`Graph`] that does nothing.
--- a/R/PipeOpNOP.R
+++ b/R/PipeOpNOP.R
@@ -60,7 +60,13 @@ PipeOpNOP = R6Class("PipeOpNOP",
   inherit = PipeOp,
   public = list(
     initialize = function(id = "nop", param_vals = list()) {
-      super$initialize(id, param_vals = param_vals,
+
+      ps = ParamSet$new(list(
+        ParamFct$new(id = "phase", default = "both", levels = c("both", "train", "predict"),
+                     tags = c("train", "predict"))
+      ))
+
+      super$initialize(id, param_set = ps, param_vals = param_vals,
         input = data.table(name = "input", train = "*", predict = "*"),
         output = data.table(name = "output", train = "*", predict = "*"),
         tags = "meta"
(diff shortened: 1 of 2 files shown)


RaphaelS1:
Not sure why this is creating a build error in PipeOpTextVectorizer... will take a proper look later

RaphaelS1:
Manually checked the build on my computer and runs fine. I don't think this PR caused the error

mb706:
what is the usecase for this?

RaphaelS1:
> what is the usecase for this?

As a practical example see `pipeline_survtoregr` (https://github.com/mlr-org/mlr3proba/blob/PipeOpTransformers/R/pipelines.R). Here the NOP is being used to pass the task only in the predict phase to `trafopred_regrsurv` which expects a task in predict but NULL in train

mb706:
Are there things that produce tasks only during predict and not during train? Or why wouldn't trafopred_regrsurv have a train-time `Task` input (which just gets ignored)?

RaphaelS1:
I discussed this with @pfistfl and in particular just adding a task that's ignored or a `*`, but that seems like a hacky fix that doesn't generalise well.

To explain why that doesn't make sense in my example, a prediction transformer should clearly do nothing in training so passing anything except NULL to this would be counter intuitive. But why during prediction should it be restricted to only taking in limited object types if it also needs a task? Here the prediction objects requires the task to get details like row ids, and more importantly the true outcome data. And the NOP is the only way (that I know / been told about) of accessing the task in the first step without before it is transformed or mutated somehow.

mb706:
I think this should not go in NOP, because then it just would not be a NOP any more. I am currently against merging this and would suggest a different solution:

Maybe a different way of looking at this problem is that `NULL` is the ultimate superclass and everything should be freely convertible to `NULL`. `mlr3pipelines` has a mechanism for automatically upcasting things (e.g. `data.table` to `data.frame`) via the [typecheck / type conversion mechanism](https://github.com/mlr-org/mlr3pipelines/blob/master/R/typecheck.R). It should only need slight tweaking to get this to work.

mb706:
Demo:
```r
library("mlr3pipelines")
library("mlr3")
library("R6")
library("data.table")
# PipeOp that just creates `tsk("iris")` in train and predict
po1 <- R6Class("test", inherit = PipeOp,
  public = list(
    initialize = function()
      super$initialize("test",
        input = data.table(name = "in", train = "*", predict = "*"),
        output = data.table(name = "out", train = "Task", predict  = "Task"))),
  private = list(
    .train = function(...) {
      self.state = list()
      list(tsk("iris"))
    },
    .predict = function(...) list(tsk("iris"))))$new()
# PipeOp that forwards its input, but it accepts only
# NULL during train and a `Task` during predict.
po2 <- R6Class("test", inherit = PipeOp,
  public = list(
    initialize = function() super$initialize("test2",
      input = data.table(name = "in", train = "NULL", predict = "Task"),
      output = data.table(name = "out", train = "*", predict = "*"))),
  private = list(
    .train = function(...) self.state = list(...),
    .predict = function(...) list(...)))$new()

# Currently we can't put these together as is
(po1 %>>% po2)$train(1)
#> Error: Output type of PipeOp test during training (Task) incompatible with input type of PipeOp test2 (NULL)

# But we can inform the system about the
# possibility of upcasting to NULL
register_autoconvert_function("NULL", function(x) {
  cat(sprintf('Autoconverting class "%s" to NULL', paste(class(x), collapse = '", "')))
  NULL
})
add_class_hierarchy_cache(c("Task", "NULL"))

(po1 %>>% po2)$train(1)
#> Autoconverting class "TaskClassif", "TaskSupervised", "Task", "R6" to NULL$test2.out
#> $test2.out$`in`
#> NULL
#>
(po1 %>>% po2)$predict(1)
#> $test2.out
#> $test2.out$`in`
#> <TaskClassif:iris> (150 x 5)
#> * Target: Species
#> * Properties: multiclass
#> * Features (4):
#>   - dbl (4): Petal.Length, Petal.Width, Sepal.Length, Sepal.Width
```

My suggestion is basically that `register_autoconvert_function` should be in the package itself and the class hierarchy should be implicit somewhere. Unless there are things that we never want to convert to NULL?

mb706:
Can you have a look if #461 works for you and this can be closed?

PR #461, opened 2020-08-05
autoconvert to NULL

mb706:
See https://github.com/mlr-org/mlr3pipelines/pull/451 and in particular [this comment and the following of mine](https://github.com/mlr-org/mlr3pipelines/pull/451#issuecomment-669523116).

If someone can add tests (e.g. from #451) and if @RaphaelS1 can confirm that this works for him this can be merged.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# mlr3pipelines 0.2.2
+
+* NULL input channels accept any kind of input
+
 # mlr3pipelines 0.2.1
 
 * New feature: Multiplicities: implicit repetition of operations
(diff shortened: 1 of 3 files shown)


RaphaelS1:
I will try and use this in a pipeline I have tomorrow that requires it and see if it works as expected for the use case.

mb706:
Tests fail because we actually have an assertion somewhere about `Task %>>% NULL` failing. The test should probably be changed to something that actually still fails.

RaphaelS1:
Thanks @mb706 works as expected, so basically as far as I'm concerned it's just like using a NOP but don't need to set 'phase' because it automatically converts to NULL?

mb706:
@sumny could you write a test or two for this and adjust the broken test?

PR #474, opened 2020-08-16
add DOT output to Graphs

### LINKED ISSUE #458: Support dot output for pipelines
Would be nice to have for markdown. See https://github.com/yihui/knitr-examples/blob/master/057-engine-dot.Rmd for an example.

sumny:
closes #458 

adds `to_dot()` method to `Graph`s which prints a DOT representation to the console. Can also be directly used in R markdown via e.g.:

`{r, dot-ex, engine = "dot", code = capture.output(g$to_dot())}`

@mllg any suggestions w.r.t adding additional arguments, e.g., for prettyfying etc.?


mllg:
LGTM for a first draft.

mb706:
The plan is to see if this can be an optional argument to `$print()` (and therefore `print(graph, ...)`)

mb706:
Otherwise its pretty nice

PR #480, opened 2020-08-28
test with testthat dev, fix some regexp in tests

### LINKED ISSUE #479: Failures with dev testthat
When I run R CMD check on mlr3pipelines with dev testthat, I see a lot of failures like:

```
Training debug.multi with input list(input_1 = 1, input_2 = 1)
-- FAILURE (test_Graph.R:221:3): assert_graph test -----------------------------
Object addresses differ at path ROOT is not TRUE

`actual`:   FALSE
`expected`: TRUE 

-- FAILURE (test_Graph.R:221:3): assert_graph test -----------------------------
Object addresses differ at path ROOT is not NULL

`actual` is a character vector ('ROOT')
`expected` is NULL

-- FAILURE (test_Graph.R:221:3): assert_graph test -----------------------------
Object addresses differ at path ROOT->[element .__enclos_env__] '.__enclos_env__' is not TRUE

`actual`:   FALSE
`expected`: TRUE 
```

Would you mind helping me figure out what's gone wrong and whether this might be a bug in testthat?

sumny:
closes #479


mb706:
can you make this work with both versions of testthat (`isn't` and `is not`)?

mb706:
merging this for cran push

PR #484, opened 2020-08-29
Fix OVR

pfistfl:
OVR had a bug, that made it not work together with resampling and the code below would fail:

```r
gr = ppl('ovr', lrn('classif.rpart'))
gr$train(tsk("iris")$filter(c(1:30, 51:80, 101:130)))
gr$predict(tsk("iris")$filter(c(1:30, 51:80, 101:130)))
```

My changes now make use of `mlr3`'s new `convert_task` function (will be merged soon) in order to copy `col_roles`, `row_roles`and so on.


pfistfl:
@mb706 could you perhaps review and merge? (PR is only 5 lines or so)

R/PipeOpOVR.R
```diff
@@ -93,15 +93,16 @@ PipeOpOVRSplit = R6Class("PipeOpOVRSplit",
       list(as.Multiplicity(private$.splittask(inputs[[1]], self$state$levels)))
     },
     .splittask = function(task, levels) {
+      new_target = paste0(task$target_names, "_tmp")
```
mb706, line 96: is it necessary to rename the target?

mb706: woudl `convert_task(ntsk$cbind(truthcol), ...` also work?

mb706: (without the rename)

pfistfl: then we'd have the same column twice in the Task.
If we would keep the same name, we would rely on the slightly strange behaviour mlr3 exhibits once two columns with the same name exist in a task (in different cbinded backends) where one column is not reachable anymore through 
`$data()`. 
So we could not rename, but this would perhaps lead to other issues.

mb706: Is overwriting columns with new data using `$cbind()` not the "official" behaviour of mlr3?

pfistfl: I don't know what is "official" there :), but I guess this could be intended behavior.
Feel free to re-write and merge. The convert_task branch will hopefully be merged tomorrow.

pfistfl:
@mb706 this is ready, the branch was merged.

PR #486, opened 2020-08-31
Bbotk in avglearner 

### LINKED ISSUE #482: Rewrite LearnerAvg with `bbotk`

pfistfl:
closes #482


mb706:
Please use one of the [github keywords for closing issues through PRs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) when referencing an issue that should obviously closed after merging.

pfistfl:
I would merge this once CI passes if there are no objections!

mb706:
please don't merge w/o review

R/LearnerAvg.R
```diff
@@ -45,20 +55,31 @@
 LearnerClassifAvg = R6Class("LearnerClassifAvg", inherit = LearnerClassif,
   public = list(
     initialize = function(id = "classif.avg") {
+      ps = ParamSet$new(params = list(
+        ParamUty$new("measure", custom_check = function(x)
+          check_r6(if (is.character(x)) msr(x) else x, "MeasureClassif"), tags = "train"),
```
mb706, line 60: (1) Since the same checks are done both here and in the RegrAvg, maybe they should be functions in the package namespace.

(2) I don't think `msr(x)` needs to be `check_r6`'d. Also this throws an error if `x` is not in `msr()`, when it only should return TRUE or a character(1). This should probably be in the direction of
```r
if (is.character(x)) {
  if (x %nin% mlr_measures$keys()) "Is a `character` but not a known measure" else TRUE
} else {
  check_r6(x, "MeasureClassif")
}
```

R/LearnerAvg.R
```diff
@@ -163,35 +191,45 @@ LearnerRegrAvg = R6Class("LearnerRegrAvg", inherit = LearnerRegr,
...
+      }))
+      optimizer = pars$optimizer
+      if (inherits(optimizer, "character")) {
+        optimizer = bbotk::opt(optimizer)
+        if (inherits(optimizer, "OptimizerNLoptr")) {
+          optimizer$param_set$values = list(xtol_rel = 1e-8, algorithm = "NLOPT_LN_COBYLA", x0 = rep(1/n_weights, n_weights))
+        }
+      }
+      measure = pars$measure
+      if (is.character(measure)) measure = msr(measure)
```
mb706, line 217: if `as_measure()` works here use that

R/LearnerAvg.R
```diff
@@ -163,35 +191,45 @@ LearnerRegrAvg = R6Class("LearnerRegrAvg", inherit = LearnerRegr,
...
+        terminator = bbotk::trm("combo", terminators = list(
+          bbotk::trm("stagnation", iters = 20 * n_weights),
+          bbotk::trm("evals", n_evals = 50 * n_weights)
+        ))
+      )
+      lgr = lgr::get_logger("bbotk")
+      old_threshold = lgr$threshold
+      lgr$set_threshold(self$param_set$values$log_level)
+      optimizer$optimize(inst)
+      on.exit(lgr$set_threshold(old_threshold))
```
mb706, line 233: this should be *before* `lgr$set_threshold(self$param_set$values$log_level)` so that things get reset if the optimizer fails (and even in the case where `self$apram_set$values$log_level` is invalid or smth.

R/LearnerAvg.R
```diff
@@ -163,35 +191,45 @@ LearnerRegrAvg = R6Class("LearnerRegrAvg", inherit = LearnerRegr,
...
+      pars = self$param_set$get_values(tags = "train")
+      ps = ParamSet$new(params = imap(data, function(x, n) {
+        if (is.numeric(n)) n = paste0("w.", n)
+        ParamDbl$new(id = n, lower = 0, upper = 1)
+      }))
+      optimizer = pars$optimizer
+      if (inherits(optimizer, "character")) {
+        optimizer = bbotk::opt(optimizer)
+        if (inherits(optimizer, "OptimizerNLoptr")) {
+          optimizer$param_set$values = list(xtol_rel = 1e-8, algorithm = "NLOPT_LN_COBYLA", x0 = rep(1/n_weights, n_weights))
```
mb706, line 213: Maybe we want to set the x0 and xtol even if the optimizer was given as an argument (unless these values were set to something else by the user?) Or is that not possible? (i.e. is `x0` a "required" parameter and the user would always have to set these parameters?)

mb706 (changes requested):

R/LearnerAvg.R
pfistfl: There is no `as_measure.character`

R/LearnerAvg.R
pfistfl: `x0` is required for some algorithms e.g. COBYLA but not for all. 
We could do this, but I do not see much added benefit and we might run into strange behavior when e.g. a different algorithm is set.

I just wanted to have this be the default as it IIRC reflects what worked best in Ledell, 2015.

R/LearnerAvg.R
pfistfl: I factored them out.

R/LearnerAvg.R
mb706: huh ok, thought this is what `as_measure` was good for.

PR #498, opened 2020-09-07
FIXME in PipeOpTaskPreproc

damirpolat:
Preprocessing operators used to remove features that were not selected from `col_roles` as described in FIXME comment. 

I fixed the problem by saving removed `col_roles` and performing union with the processed columns.

Thanks,
Damir


damirpolat:
Not sure why checks are failing. Everything is fine when I run `devtools:check()` locally. Unless I'm missing something?

mb706:
Has now been fixed in #866 . Sorry for never coming back about this.

PR #499, opened 2020-09-08
Fixes for mlr3#539 (container refactoring)

mllg:
For https://github.com/mlr-org/mlr3/pull/539.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -84,7 +84,7 @@ Suggests:
 RdMacros:
     mlr3misc
 Remotes:
-    mlr-org/mlr3,
+    mlr-org/mlr3@refactor_containers3,
     Bioconductor/Biobase
 ByteCompile: true
 Encoding: UTF-8
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,8 +47,8 @@ Imports:
     data.table,
     digest,
     lgr,
-    mlr3 (>= 0.1.4),
-    mlr3misc (>= 0.1.4),
+    mlr3 (>= 0.6.0),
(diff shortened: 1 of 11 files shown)


pfistfl:
Looks like the tests were actually wrong before, but it did not pop up.
If I understand this correctly,

```r
tt$param_set$values$targetmutate.trafo = function(x) exp(x)
 tt$param_set$values$targetmutate.inverter = function(x) log(x)
```

should be 

```r
tt$param_set$values$targetmutate.trafo = function(x) exp(x$response)
 tt$param_set$values$targetmutate.inverter = function(x) log(x$response)
```

in L53:54 of https://github.com/mlr-org/mlr3pipelines/blob/master/tests/testthat/test_mlr_graphs_targettrafo.R

sumny:
> Looks like the tests were actually wrong before, but it did not pop up.
> If I understand this correctly,
> 
> ```r
> tt$param_set$values$targetmutate.trafo = function(x) exp(x)
>  tt$param_set$values$targetmutate.inverter = function(x) log(x)
> ```
> 
> should be
> 
> ```r
> tt$param_set$values$targetmutate.trafo = function(x) exp(x)
>  tt$param_set$values$targetmutate.inverter = function(x) list(response = log(x$response))
> ```
> 
> in L53:54 of https://github.com/mlr-org/mlr3pipelines/blob/master/tests/testthat/test_mlr_graphs_targettrafo.R

yes, this is something I did not catch everywhere when integrating the new trafo changes. Apparently, the tests did not fail because within the old `prediction$data$tab` slot, this automatically was recycled and somehow worked..

sumny (approved):

PR #513, opened 2020-10-01
open up PipeOpLearnerCV to all resampling methods

### LINKED ISSUE #500: Extend PipeOpLearnerCV for other resamplings
`levels = c("cv", "insample")` is currently very limited and not flexibly extensible i.e. for mlr3spatiotmpcv resamplings.
Might need some mechanism to allow a subset of mlr_resamplings

sumny:
Allow all resamplings currently listed in `mlr_resamplings`. Closes #500 

If a resampling returns multiple predictions for a row id, then they are aggregated using the mean; if `task_type = "classif"` aggregation for `response` is done using the mode; maybe we should simply base this on the argmax of the mean aggregation of the probs instead - if available). Maybe we also should open up to custom aggregation functions that can be passed as a hyperparameter but we have to make sure that boundaries of probs etc. are respected, e.g., all aggregated probs must be within `[0, 1]` and the sum for one row id must still be `1` (or we enforce this later),

If a resampling fails to return predictions for a row id present in the input task, this row id is added with missing values.
All in all, this results in the row ids of the input task matching the row ids of the output features based on the resampled prediction.

For custom resampling, `train_sets` and `test_sets` are currently passed as `ParamUty`s `"resampling.custom.train_sets"` and `"resampling.custom.test_sets"`; not sure if this is the best way here.

### DIFF
--- a/R/PipeOpLearnerCV.R
+++ b/R/PipeOpLearnerCV.R
@@ -110,6 +110,7 @@
 #' graph$pipeops$classif.rpart$learner$predict_type = "prob"
 #'
 #' graph$train(task)
+# FIXME: docs and tests
 PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
   inherit = PipeOpTaskPreproc,
   public = list(
@@ -121,11 +122,15 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
       task_type = mlr_reflections$task_types[get("type") == private$.learner$task_type][order(get("package"))][1L]$task
 
       private$.crossval_param_set = ParamSet$new(params = list(
-        ParamFct$new("method", levels = c("cv", "insample"), tags = c("train", "required")),
+        ParamFct$new("method", levels = c("bootstrap", "custom", "cv", "holdout", "insample", "loo", "repeated_cv", "subsampling"), tags = c("train", "required")),
+        ParamInt$new("repeats", lower = 1L, tags = c("train", "required")),
         ParamInt$new("folds", lower = 2L, upper = Inf, tags = c("train", "required")),
-        ParamLgl$new("keep_response", tags = c("train", "required"))
+        ParamDbl$new("ratio", lower = 0, upper = 1, tags = c("train", "required")),
+        ParamLgl$new("keep_response", tags = c("train", "required")),
+        ParamUty$new("train_sets", tags = "train", custom_check = function(x) check_list(types = "atomicvector", any.missing = FALSE)),
+        ParamUty$new("test_sets", tags = "train", custom_check = function(x) check_list(types = "atomicvector", any.missing = FALSE))
       ))
-      private$.crossval_param_set$values = list(method = "cv", folds = 3, keep_response = FALSE)
+      private$.crossval_param_set$values = list(method = "cv", repeats = 30L, folds = 3, ratio = 2 / 3, keep_response = FALSE)
       private$.crossval_param_set$set_id = "resampling"
       # Dependencies in paradox have been broken from the start and this is known since at least a year:
       # https://github.com/mlr-org/paradox/issues/216
@@ -169,14 +174,45 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
       self$state = private$.learner$train(task)$state
       pv = private$.crossval_param_set$values
 
-      # Compute CV Predictions
-      if (pv$method != "insample") {
-        rdesc = mlr_resamplings$get(pv$method)
-        if (pv$method == "cv") rdesc$param_set$values = list(folds = pv$folds)
-        rr = resample(task, private$.learner, rdesc)
-        prds = as.data.table(rr$prediction(predict_sets = "test"))
-      } else {
-        prds = as.data.table(private$.learner$predict(task))
+      if (pv$method == "insample") {
+        return(private$pred_to_task(as.data.table(private$.learner$predict(task)), task))  # early exit
+      }
+
+      # Compute resampled Predictions
+      rdesc = mlr_resamplings$get(pv$method)
+      rdesc$param_set$values = switch(pv$method,
+        "bootstrap" = list(repeats = pv$repeats, ratio = pv$ratio),
+        "custom" = list(),
+        "cv" = list(folds = pv$folds),
+        "holdout" = list(ratio = pv$ratio),
+        "loo" = list(),
+        "repeated_cv" = list(repeats = pv$repeats, folds = pv$folds),
+        "subsampling" = list(repeats = pv$repeats, ratio = pv$ratio))
+      if (pv$method == "custom") {
+        rdesc$instantiate(task, train_sets = private$.crossval_param_set$values$train_sets, test_sets = private$.crossval_param_set$values$test_sets)
+      }
+      rr = resample(task, private$.learner, rdesc)
+      prds = as.data.table(rr$prediction(predict_sets = "test"))
+      nrows_duplicated = length(prds$row_id[duplicated(prds$row_id)])
+      missing_rows = setdiff(task$row_ids, prds$row_id)
+      nrows_missing = length(setdiff(task$row_ids, prds$row_id))
+
+      if (nrows_duplicated || nrows_missing) {  # duplicates or missings
+        SDcols = setdiff(colnames(prds), c("row_id", "truth"))
+        prds_corrected = if (nrows_duplicated) {
+          prds[, map(.SD, aggregation), by = "row_id", .SDcols = SDcols]
+        } else {
+          setNames(data.table(matrix(nrow = 0L, ncol = NCOL(prds))), colnames(prds))
+        }
+        prds_extended = as.list(prds_corrected)[SDcols]
+        prds_extended = map(prds_extended, add_missings, len = nrows_missing)
+        prds_extended[["row_id"]] = c(prds_corrected[["row_id"]], missing_rows)
+        prds = setDT(prds_extended)
+
+        target = task$truth(prds$row_id)
+        if (task$task_type == "classif") {
+          prds$response = factor(prds$response, levels = levels(target), ordered = is.ordered(target))
+        }
       }
 
       private$pred_to_task(prds, task)
@@ -203,5 +239,28 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
     .learner = NULL
   )
 )
+
+# Helper function for aggregating predictions if duplicated rows are present:
+#  - handles response, prob etc. naturally
+#  - if x is a factor (e.g., response if classif) take the mode and return this level as a character (factor fix is applied later)
+#  - if x is numeric (e.g., response if regr, or prob or se), take the mean (for prob this is invariant w.r.t to [0, 1] boundaries)
+aggregation = function(x) {
+  if (length(x) == 1L) {
+    return(x)  # early exit
+  }
+  if (is.factor(x)) {
+    tt = table(x)
+    names(tt[which.max(tt)])
+  } else {
+    mean(x, na.rm = TRUE)
+  }
+}
+
+# Helper function to add missings to predictions based on their storage mode
+add_missings = function(x, len) {
+  c(x, switch(typeof(x),
+    "character" = rep_len(NA_character_, length.out = len),
+    "double" = rep_len(NA_real_, length.out = len)))
+}
 
 mlr_pipeops$add("learner_cv", PipeOpLearnerCV, list(R6Class("Learner", public = list(id = "learner_cv", task_type = "classif", param_set = ParamSet$new()))$new()))
--- a/R/PipeOpLearnerCV.R
+++ b/R/PipeOpLearnerCV.R
@@ -7,7 +7,7 @@
 #' @description
 #' Wraps an [`mlr3::Learner`] into a [`PipeOp`].
 #'
-#' Returns cross-validated predictions during training as a [`Task`][mlr3::Task] and stores a model of the
+#' Returns resampled predictions during training as a [`Task`][mlr3::Task] and stores a model of the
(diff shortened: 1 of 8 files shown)


R/PipeOpLearnerCV.R
```diff
@@ -43,7 +42,7 @@
 #' type given to `learner` during construction; both during training and prediction.
 #'
 #' The output is a task with the same target as the input task, with features replaced by predictions made by the [`Learner`][mlr3::Learner].
-#' During training, this prediction is the out-of-sample prediction made by [`resample`][mlr3::resample], during prediction, this is the
+#' During training, this prediction is the prediction made by [`resample`][mlr3::resample], during prediction, this is the
```
pfistfl, line 45: I would stick to out-of-sample.

R/PipeOpLearnerCV.R
```diff
@@ -64,10 +63,24 @@
 #' The parameters are the parameters inherited from the [`PipeOpTaskPreproc`], as well as the parameters of the [`Learner`][mlr3::Learner] wrapped by this object.
 #' Besides that, parameters introduced are:
 #' * `resampling.method` :: `character(1)`\cr
-#'   Which resampling method do we want to use. Currently only supports `"cv"` and `"insample"`. `"insample"` generates
-#'   predictions with the model trained on all training data.
-#' * `resampling.folds` :: `numeric(1)`\cr
-#'   Number of cross validation folds. Initialized to 3. Only used for `resampling.method = "cv"`.
+#'   Which resampling method to use. Supports `"cv"`,`"bootstrap"`, `"holdout"`, `"loo"`, `"repeated_cv"`, `"subsampling"`, `"custom"` and `"insample"`.
```
pfistfl, line 66: I would not list all here, instead reference `mlr_resamplings` and mention that this is extended by various packages.

R/PipeOpLearnerCV.R
```diff
@@ -64,10 +63,24 @@
...
+#'   See [`mlr_resamplings`][mlr3::mlr_resamplings].
+#'   `"insample"` generates predictions with the model trained on all training data.
+#'   In the case of the resampling method returing multiple predictions per row id, the predictions are aggregated via their mean
+#'   (execpt for the `"response"` in the case of a [classification Task][mlr3::TaskClassif] which is aggregated using the mode).
+#'   In the case of the resampling method not returning predictions for all row ids as given in the input [`Task`][mlr3::Task], these predictions are added as missing.
+#' * `resampling.repeats` :: `integer(1)`\cr
```
pfistfl, line 72: Do we really want to have all args here? I feel this does not scale considering future extensions (spatiotempcv etc.)

sumny:
I guess if we want to be really flexible with respect to which `Resampling`s to include, there is no other way than also wrapping the `Resampling` as provided by the user. The current status of this PR is exactly this, i.e., we not only wrap the `Learner` but also the `Resampling` (initialized to `rsmp("cv", folds = 3)`, e.g. as before).

Test currently fail because:
* There are issues with deep cloning `param_sets` of `Resampling`s
* The parameters of `Resampling`s do neither have a train nor predict tag (#510 would allow for skipping this test here)

sumny:
Some comments: Ideally we do not aggregate / deal with multiple (or none) predictions per `row_id` but rather have a `PipeOpAggregate` or something that operates on some meta info of the output `Task`

sumny:
some more comments: add `row_reference` to `task_col_roles`. `LearnerCV` then simply appends the row reference to the task and the aggregator/`PipeOpJoin` then handles it

sumny:
I just did a reiterate which I summarize here:
- `PipeOpLearnerCV` now wraps any `Resampling` and simply returns the predictions. Due to the fact that some `Resampling ` methods may return multiple predictions per original row id, it also append a new column of `col_role` `row_reference` to the data
- Based on such a `row_reference` column one can then use `PipeOpAggregate` to aggregate these multiple predictions per original row id
- There is still the "problem" that some `Resampling` methods must not necessarily return a prediction for each original row id, i..e, in subsampling the returned predictions (w.r.t to row ids) are only a subset of the original row ids; in this case stacking via featureunion is not possible but I guess this is fine (#410 may fix this in the future)

R/PipeOpLearnerCV.R
```diff
@@ -159,53 +165,71 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
...
-      if (pv$method != "insample") {
-        rdesc = mlr_resamplings$get(pv$method)
-        if (pv$method == "cv") rdesc$param_set$values = list(folds = pv$folds)
-        rr = resample(task, private$.learner, rdesc)
-        prds = as.data.table(rr$prediction(predict_sets = "test"))
-      } else {
-        prds = as.data.table(private$.learner$predict(task))
-      }
+      # compute resampled predictions
+      rr = resample(task, private$.learner, private$.resampling)
```
mb706, line 185: is there a way to check if the resampling fits a model on all the data, which could then be used for prediction without needing to fit twice?

R/PipeOpLearnerCV.R
```diff
@@ -113,29 +119,29 @@
...
-      private$.crossval_param_set$values = list(method = "cv", folds = 3, keep_response = FALSE)
-      private$.crossval_param_set$set_id = "resampling"
-      # Dependencies in paradox have been broken from the start and this is known since at least a year:
-      # https://github.com/mlr-org/paradox/issues/216
-      # The following would make it _impossible_ to set "method" to "insample", because then "folds"
-      # is both _required_ (required tag above) and at the same time must be unset (because of this
-      # dependency). We will opt for the least annoying behaviour here and just not use dependencies
-      # in PipeOp ParamSets.
-      # private$.crossval_param_set$add_dep("folds", "method", CondEqual$new("cv"))  # don't do this.
+      private$.additional_param_set$values = list(keep_response = FALSE)
```
mb706, line 140: name `resampling.keep_response` (and hope no resampling method has `keep_response` as parameter)

R/PipeOpLearnerCV.R
```diff
@@ -159,53 +165,71 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
...
+      backend = if (identical(sort(prds[[row_reference]]), sort(task$row_ids))) {
+        set(prds, j = task$backend$primary_key, value = prds[[row_reference]])
+        as_data_backend(prds, primary_key = task$backend$primary_key)
+      } else {
+        as_data_backend(prds)
+      }
+
+      # get task_type from mlr_reflections and call constructor
+      constructor = get(mlr_reflections$task_types[["task"]][chmatch(task$task_type, table = mlr_reflections$task_types[["type"]], nomatch = 0L)][[1L]])
+      newtask = invoke(constructor$new, id = task$id, backend = backend, target = task$target_names, .args = task$extra_args)
```
mb706, line 222: Note to @mb706, this needs to be brought in accord with PipeOpTaskPreproc's `affect_columns`.

R/PipeOpLearnerCV.R
```diff
@@ -159,53 +165,71 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
...
+    .pred_to_task = function(prds, task) {
+      if (!self$param_set$values$keep_response && self$learner$predict_type == "prob") {
         prds[, response := NULL]
       }
-      renaming = setdiff(colnames(prds), c("row_id", "row_ids"))
-      setnames(prds, renaming, sprintf("%s.%s", self$id, renaming))
+      renaming = setdiff(colnames(prds), c("row_ids", "truth"))
+      setnames(prds, old = renaming, new = sprintf("%s.%s", self$id, renaming))
+      setnames(prds, old = "truth", new = task$target_names)
+      row_reference = paste0("pre.", self$id)
```
mb706, line 205: change

R/PipeOpLearnerCV.R
```diff
@@ -159,53 +165,71 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
...
+      while (row_reference %in% task$col_info$id) {
+        row_reference = paste0(row_reference, ".")
```
mb706, line 207: instead throw an error when IDs collide, user has to change PipOp's ID then. In any case IDs are unique within graph so usually shouldn't be a problem.

R/PipeOpLearnerCV.R
```diff
@@ -159,53 +165,71 @@ PipeOpLearnerCV = R6Class("PipeOpLearnerCV",
...
 
-      # This can be simplified for mlr3 >= 0.11.0;
-      # will be always "row_ids"
-      row_id_col = intersect(colnames(prds), c("row_id", "row_ids"))
-      setnames(prds, old = row_id_col, new = task$backend$primary_key)
-      task$select(character(0))$cbind(prds)
+      # the following is needed to pertain correct row ids in the case of e.g. cv
+      # here we do not necessarily apply PipeOpAggregate later
+      backend = if (identical(sort(prds[[row_reference]]), sort(task$row_ids))) {
+        set(prds, j = task$backend$primary_key, value = prds[[row_reference]])
```
mb706, line 214: check primary_key is not in names(prds)

mb706: (assert)

R/PipeOpLearnerCV.R
mb706: But it should be the PipeOpTaskPreproc's responsibility to keep all col-roles (that are not disabled by affect_columns), so in particular should respect weights etc.

Maybe use [this](https://github.com/mlr-org/miesmuschel/blob/188b2ad70e6e563ab3ee018fd8e94ceacbd5bc89/R/PipeOpDensitySplit.R#L111-L115) for inspiration.

mb706: Also the previous-id-column should really be a different col-role to avoid accidentally training on the ID.

R/PipeOpTuneThreshold.R
```diff
@@ -143,7 +143,12 @@ PipeOpTuneThreshold = R6Class("PipeOpTuneThreshold",
     },
     .task_to_prediction = function(input) {
       prob = as.matrix(input$data(cols = input$feature_names))
-      colnames(prob) = unlist(input$levels())
+      # setting the column names the following way is safer
+      nms = map_chr(strsplit(colnames(prob), "\\."), function(x) x[length(x)])
```
mb706, line 147: maybe breaks when factor level has a period?

mb706: Better way: use input$levels(input$target_name), generate putative colnames from that, then compare to given col names. the assert is good though.

R/PipeOpTuneThreshold.R
```diff
@@ -143,7 +143,12 @@ PipeOpTuneThreshold = R6Class("PipeOpTuneThreshold",
     },
     .task_to_prediction = function(input) {
       prob = as.matrix(input$data(cols = input$feature_names))
-      colnames(prob) = unlist(input$levels())
```
mb706, line 146: this worked before because there was only one col in the task with factors for classification, but it is very brittle and should be avoided.

R/zzz.R
```diff
@@ -15,6 +15,9 @@ register_mlr3 = function() {
     c("abstract", "meta", "missings", "feature selection", "imbalanced data",
     "data transform", "target transform", "ensemble", "robustify", "learner", "encode",
      "multiplicity")))
+  if (!all(grepl("row_reference", x$task_col_roles))) {
+    x$task_col_roles = map(x$task_col_roles, function(col_roles) c(col_roles, "row_reference"))
```
mb706, line 19: this doesn't work if other tasks get added after mlr3pipelines, so we should ask @mllg (1) is there a way to do things like this well?, (2) can we just have the col role like that

R/PipeOpAggregate.R
```diff
@@ -0,0 +1,148 @@
+#' @title Aggregate Features Row-Wise
+#'
+#' @usage NULL
+#' @name mlr_pipeops_aggregate
+#' @format [`R6Class`] object inheriting from [`PipeOpTaskPreprocSimple`]/[`PipeOpTaskPreproc`]/[`PipeOp`].
+#'
+#' @description
+#' Aggregates features row-wise based on multiple observations indicated via a column of role `row_reference` according to expressions given as formulas.
```
mb706, line 8: maybe we don't need to restrict to that colrole?

R/PipeOpAggregate.R
```diff
@@ -0,0 +1,148 @@
...
+#' @section Parameters:
+#' The parameters are the parameters inherited from [`PipeOpTaskPreproc`], as well as:
+#' * `aggregation` :: named `list` of `formula`\cr
+#'   Expressions for how features should be aggregated, in the form of `formula`.
+#'   Each element of the list is a `formula` with the name of the element naming the feature to aggregate and the formula expression determining the result.
+#'   Each formula is evaluated within [`data.table`] environments of the [`Task`][mlr3::Task] that contain all features split via the `by` argument (see below).
+#'   Initialized to `list()`, i.e., no aggregation is performed.
+#' * `by` :: `character(1)` | `NULL`\cr
+#'   Column indicating the `row_reference` column of the [`Task`][mlr3::Task] that should be the row-wise basis for the aggregation.
+#'   Initialized to `NULL`, i.e., no aggregation is performed.
```
mb706, line 38: Maybe also `.SDcols`

R/PipeOpAggregate.R
```diff
@@ -0,0 +1,148 @@
...
+        ParamUty$new("by", tags = c("train", "predict", "required"), custom_check = function(x) check_string(x, null.ok = TRUE))
+      ))
+      ps$values = list(aggregation = list(), by = NULL)
+      super$initialize(id, ps, param_vals = param_vals, tags = "ensemble")
+    }
+  ),
+  private = list(
+    .transform = function(task) {
+
+      if (length(self$param_set$values$aggregation) == 0L || is.null(self$param_set$values$by)) {
```
mb706, line 102: empty aggregation should not be allowed

PR #515, opened 2020-10-05
use multiplicities for bagging, update docs and tests

### LINKED ISSUE #477: Maybe ppl bagging should use multiplicities
Large graphs can be pretty slow and hard to handle (e.g. in https://github.com/mlr-org/paradox/issues/303), and there is not much reason not to use multiplicities as long as the bagging learners are all the same PipeOp

sumny:
closes #477 (related: mlr-org/paradox#303)

Old memory and time usage:
```
profvis({GraphLearner$new(ppl("bagging", lrn("regr.featureless"), iterations = 50, averager = po("regravg")))})
```

image

New memory and time usage:
```
profvis({GraphLearner$new(ppl("bagging", lrn("regr.featureless"), iterations = 50, averager = po("regravg", collect_multiplicity = TRUE)))})
```
image

Maybe we should also update the bagging section of the book soon / the gallery should be unaffected.


R/mlr_graphs_elements.R
```diff
@@ -126,28 +127,29 @@ mlr_graphs$add("robustify", pipeline_robustify)
...
 pipeline_bagging = function(graph, iterations = 10, frac = 0.7, averager = NULL) {
+  g = as_graph(graph, clone = TRUE)
   assert_count(iterations)
   assert_number(frac, lower = 0, upper = 1)
-  graph = as_graph(graph)
-  if (!is.null(averager)) averager = as_graph(averager)
-
-  subs = po("subsample", param_vals = list(frac = frac)) %>>% graph
-  subs_repls = pipeline_greplicate(subs, iterations)
+  if (!is.null(averager)) averager = as_graph(averager, clone = TRUE)
```
mb706, line 147: you should add a test that the averager does collect multiplicities (length of `input` is 1 and its train and predict type both start with a `[`).

mb706 (changes requested):

R/mlr_graphs_elements.R
mb706: also test that creating a bagging pipeline with a pipeop that does not collect multiplicities throws an error.

sumny: yes, makes sense! thanks for the heads-up!

PR #517, opened 2020-10-08
fix bug in cbind_tasks, update tests

sumny:
somewhat related: mlr-org/mlr3gallery#77

in `cbind_tasks` explicitly return `NULL` within the for-loop before building the sum when checking for real duplicates - otherwise it may fail when tasks do not share the same features.

apparently we also no longer can rely on `task$cbind()` automatically merging duplicated columns (by their names) into a single one, therefore we have to explicitly subset to the unique feature names before cbinding to the task (otherwise it will complain - at least for a `data.table` backend)


mb706:
> apparently we also no longer can rely on task$cbind() automatically merging duplicated columns (by their names

why not?

PR #549, opened 2020-11-19
make colapply inherit from PipeOpTaskPrepreocSimple

### LINKED ISSUE #539: Maybe turn PipeOpColApply into a PipeOpTaskPreprocSimple
is this still necessary? https://github.com/mlr-org/mlr3pipelines/blob/19a19b9c27286e1aea09245a0d1c5d52991e4b0a/R/PipeOpColApply.R#L52-L55 
Maybe the necessary change is (or should be) in PipeOpTaskPreprocSimple.

sumny:
closes #539 
apparently the issue did no longer persist?

### DIFF
--- a/R/PipeOpColApply.R
+++ b/R/PipeOpColApply.R
@@ -2,11 +2,11 @@
 #'
 #' @usage NULL
 #' @name mlr_pipeops_colapply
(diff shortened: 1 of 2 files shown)


mb706:
I think this was because empty datasets were making trouble, but that was fixed upstream at some point.

PR #574, opened 2021-03-22
test with paradox performance improvement

mb706:
see if tests pass with proposed changes in paradox


mb706:
not needed any more

PR #580, opened 2021-05-19
Fix some tests for upcoming mlr3

mllg:


mllg:
@mb706 can you merge this and do a release soonish?

PR #588, opened 2021-07-14
Test against mlr3devel

mllg:


.github/workflows/tic.yml
```diff
@@ -18,15 +18,16 @@ jobs:
   all:
     runs-on: ${{ matrix.config.os }}
 
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+    name: ${{ matrix.config.os }} - r-${{ matrix.config.r }} - mlr3-${{ matrix.config.mlr3 }}
```
pat-s, line 21: I don't think you need to change the matrix naming and cache, conditioning on an env var as in the `config:` and then adding the if-condition as you did in `tic.R` should be enough.

.github/workflows/tic.yml
```diff
@@ -18,15 +18,16 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
         config:
           # use a different tic template type if you do not want to build on all listed platforms
-          - { os: ubuntu-latest, r: "devel" }
-          - { os: ubuntu-latest, r: "release", pkgdown: true, latex: true }
+          - { os: ubuntu-latest, r: "devel", mlr3: "release"}
+          - { os: ubuntu-latest, r: "release", mlr3: "release", pkgdown: true, latex: true }
+          - { os: ubuntu-latest, r: "release", mlr3: "devel" }
```
pat-s, line 30: This works and is explicit, you could also just set one for the devel runner like `devel: "true"` and then use `ci_has_env("devel")` in `tic.R`.

PR #591, opened 2021-08-17
update docs for pipeop histbin

### LINKED ISSUE #584: PipeOpHistBin param docs and implementation missmatch
The docs indicate `hist` while implementation has a `breaks` argument

pfistfl:
fixes #584

We called the argument `bins` while `hist()` called it `breaks`.
I think we should stick with the original names in order to avoid confusion when we want users to specify args 
that for a specific function.


mb706:
maybe tests need to be adjusted?

mb706:
Fixed the tests and merged in https://github.com/mlr-org/mlr3pipelines/pull/612

PR #624, opened 2021-10-02
Pofu

mb706:
Try to solve some issues under the [POFU Tag](https://github.com/mlr-org/mlr3pipelines/issues?q=is%3Aissue+is%3Aopen+label%3A%22Tag%3A+POFU%22)


mb706:
See also https://github.com/mlr-org/mlr3/issues/942

PR #630, opened 2021-11-09
Prepare new cran submission

mb706:


mllg:
Have you uploaded this version?

pat-s:
According to https://lockedata.github.io/cransays/articles/dashboard.html, yes.

However, I don't understand why this is "patch" release when looking at the release notes which include three bullet point items starting with "New [...]".

mb706:
Fair point, I bunped the version to 0.4

PR #633, opened 2021-11-28
Fix fallback of predict_type in GraphLearner

mllg:
If the heuristic in `GraphLearner` cannot determine the `predict_type`, it defaults to all available predict types stored in `mlr_reflections` for the first predict type setting. In `mlr3proba` this leads to `predict_type` being a `character(4)` instead of `character(1)` which leads to further problems down the line.

```r
library(mlr3proba)
as_learner(mlr3pipelines::ppl("distrcompositor", learner = lrn("surv.coxph")))$predict_type
```

This PR fixes this behavior, setting the `predict_type` to the first predict type setting.


mllg:
ping @mb706

PR #639, opened 2022-01-20
Do not call set.seed() in tests

mllg:
Otherwise everything gets deterministic (unless executed in parallel).


mb706:
closing this since I never got around thinking if this is useful, and I will probably never get around to it. It now has quite a few concepts already; maybe at some point we will do this as part of a general refactor of the tests.

PR #640, opened 2022-01-21
Test different method to source helper

mllg:


mllg:
Loading mlr3 before sourcing the helper files avoids a namespace problem with the current release. This should repair the CI.

PR #654, opened 2022-03-04
feat: add label and man field

be-marc:


mllg:
Can we merge this? Would be nice for the homepage to have some labels...

PR #673, opened 2022-07-05
fix: Connecting a channel of a multi-output pipeop

sebffischer:


sebffischer:
Before, it was impossible to connect a channel of a PipeOp with > 1 output channel with > 1 input channel of a different PipeOp. Subsetting using `..output[get("src_id")]` recycles the list correspondingly

PR #702, opened 2022-09-30
feature: allow id suffixes when retrieving dictionary element

sebffischer:


sebffischer:
This could also be added to `mlr3misc`, but maybe first add it here so we don't have to wait for a mlr3misc release to have this in mlr3pipelines?

R/utils.R
```diff
@@ -103,3 +103,63 @@ multiplicity_recurse = function(.multip, .fun, ...) {
...
+#' dictionary_sugar_inc_mget(mlr_pipeops, c("pca_1", "pca_2"))
+#'
+#' @rdname inc_get
+#' @export
+dictionary_sugar_inc_get = function(dict, .key, ...) {
+  newkey = gsub("_\\d+$", "", .key)
+  add_suffix = .key != newkey
+  if (add_suffix) {
+    assert_true(!methods::hasArg("id"))
+    suffix = gsub(newkey, "", .key)
```
mb706, line 141: I think its usually not a good idea to use a user-supplied string as regex, since strictly speaking you'd have to escape regex special characters. I'd suggest
```r
regmatches(.key, regexpr("_\\d+$", key))
```

mb706 (changes requested):
Looks mostly good to me. I marked one nitpick. I also don't think the functions should be exported from mlr3pipelines. I agree they should go to mlr3misc in the long run.

PR #704, opened 2022-10-13
feat: add learners conversion for pipeop and graph

sebffischer:


mb706:
I wasn't aware of `as_learners`, but if this is the intended behaviour, wouldn't it make more sense to have `as_learners.default` just call `as_learner` in every case? (see https://github.com/mlr-org/mlr3/issues/865)

PR #710, opened 2023-01-30
base_learner

### LINKED ISSUE #603: Implement base_learner method for GraphLearners
To simplify accessing fields in simple (linear) pipelines.

https://github.com/mlr-org/mlr3/issues/653

### LINKED ISSUE #699: GraphLearner$get_last_learner or something
this doesn't work in *all* cases, e.g. when doing bagging, but often it is what one wants: get the last `PipeOp` in a sequential `Graph` that is a `PipeOpLearner`, and return its (trained) `Learner`.

mb706:
closes #699 
closes #603 

@be-marc  can you check whether this works with the tooling that relies on it?

@mllg I am a it confused by how this is supposed to be implemented. [The implementation in `Learner`](https://github.com/mlr-org/mlr3/blob/222b8a98504a0deafc4db490b0f5e2401248babc/R/Learner.R#L366-L371) seems to suggest that `private$.base_learner` should be implemented, but `$base_learner()` doesn't do anything except delegate there, not even `if (recursive <= 0) return(self)` or `assert(check_int(recursive), check_numeric(recursive, lower = Inf))`.
Consequently, [`AutoTuner`](https://github.com/mlr-org/mlr3tuning/blob/b0298023e707b7d80e68254ab41fdb2bc68bd146/R/AutoTuner.R#L164-L166) just overwrites the method and does not use `private$.base_learner`. I also find it weird that `recursive = 0` does not return `self` there.


be-marc:
Looks good. The use cases that have come up for me work. Yes this should be `self` instead of `self$learner`. Thanks!

mb706:
ok, updated this PR to also have `recursive <= 0` return `self`.

PR #750, opened 2023-12-14
feat: enable test set usage in graph

sebffischer:
not quite done


sebffischer:
How do we deal with the problems where the same rows are used for training and predicting, but they are preprocessed differently? In this scenario, we cannot simply pass ONE task to the learner. 

We can deal with this problem in different ways: 
1) We can simply ensure that the test rows are disjunct from the predict rows (seems simple enough). 
2) We can make the private `$.train` method of the learner accept an additional argument `task_predict`. 

1) Would probably be easier and good enough.

sebffischer:
Also: task's rbind does not allow overwriting already existing rows it seems like

Issue #706, opened 2022-10-19
predict on test set rows


mb706:
go through each pipeop and let it predict directly, since some pipeops need the predicted set

mb706:
* PipeOpTaskPreproc checks if there are test set rows, if so calls .predict_task
* tests should verify that all pipeops that have `"Task"` inputs and outputs handle this correctly: `$predict()` and `$train()[test set rows, ]` are the same. The autotests have some flag that indicates if `$predict()` is random? in `expect_datapreproc_pipeop_class`
* some pipeops that do task preprocessing but dont inherit from pipeoptaskpreproc are worrying here: pipeopimpute. It should probably just inherit from pipeoptaskpreproc.
* PipeOpTaskPreproc *always* handles test rows if they are present. GraphLearner removes test rows in most cases, unless a PipeOp needs them. Determines this by checking each PipeOp: Does it contain a `$uses_test_rows` active binding / field? (choose not to just look for attached `$learner`, if present. Problem here: e.g. ImputeLearner should not forward its 'needs-test-set' since it *can not be used*)
* GraphLearner can set its own relevant property

sebffischer:
There is a memory inefficiency with the suggested approach: 

Let's say we have a task with data of the form 


| row_id | y | x1 |
| -------| - | ---|
|     1      |             |
| ...        |               |
|   n_use | | 
|   n_use + 1 | | 
| ...        |               |
|   n_use + n_test | | 

Lets say we apply two preprocessing operations to `x1` in two subsequent `PipeOp`s.

In the approach we discussed, we first `task$cbind()` the preprocessed train rows (`pipeop$.train_task()`and then `task$rbind()` the preprocessed test rows, resulting from `pipeop$.predict_task()`.
The problem with this is now, that the `task$rbind()` of the test rows, happens in both preprocessing PipeOps.
This means in both PipeOps, a table of the form below is rbinded. 

| row_id | y | x1 |
| -------| - | ---|
|   1 | | 
| ...        |               |
|   n_test | | 

The wasteful thing here is that we rbind column `y`, as it contains information that is simply already present in the tasks's backend and is now being duplicated. This problem comes worse when there are many columns that are left untouched by a preprocessing pipeop.

The virtual backend that would result of applying two preprocessing operations on `x1` would internally store `n_use + n_test + n_test * n_pipeop` rows, which is not so great.
[...]

PR #753, opened 2024-01-24
start working on validation support

sebffischer:
* [ ] The GraphLearner / Graph should null the test roles unless there is a PipeOp that makes use of them. Otherwise we might end up making predictions that are not needed. Maybe we want to achieve this differently / change this in `mlr3`?

PipeOps to look at: 
* [ ] PipeOpChunk
* [ ] PipeOpFeatureUnion
* [ ] PipeOpImpute
* [ ] PipeOpOVRSplit
* [x] PipeOpTuneThreshold (not really anything to do)
* [x] PipeOpLearner (not really anything to do)
* [ ] PipeOpLearnerCV (not a lot to do)
* [ ] PipeOpTrafo
* [ ] PipeOpTaskPreproc: train and predict stage now need to work a little differently even for stuff like PipeOpPreprocSimple

### DIFF
--- a/R/PipeOpTaskPreproc.R
+++ b/R/PipeOpTaskPreproc.R
@@ -72,7 +72,7 @@
 #' the [`Task`][mlr3::Task] type is the same as for input; both during training and prediction.
 #'
 #' The output [`Task`][mlr3::Task] is the modified input [`Task`][mlr3::Task] according to the overloaded
-#' `private$.train_task()`/`private$.predict_taks()` or `private$.train_dt()`/`private$.predict_dt()` functions.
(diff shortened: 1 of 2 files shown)


sebffischer:
This does not work because of the same reason bootstrapping does not work with `mlr3pipelines`. 
Consider the case where a task has overlapping "use" and "test" rows, e.g. `task$row_roles$use` of `c(1, 2)` and `task$row_roles$test` and this task should be preprocessed by `PipeOpTaskPreproc`.

We want to treat the row `2` during train and `2` during predict differently, as we want to apply the training preprocessing to the former and the prediction preprocessing to the latter. This does not work, because the tasks's data backend has only 3 rows and we want to cbind a backend that has 4 rows.

mb706:
https://github.com/mlr-org/mlr3pipelines/issues/706

PR #791, opened 2024-08-06
PipeOpUMAP

advieser:
This implements Uniform Manifold Approximation and Projection (UMAP) from the [uwot](https://cran.r-project.org/web/packages/uwot/index.html) package.

Training works via `uwot::umap2()` and prediction through `uwot::umap_transform()`.

closes https://github.com/mlr-org/mlr3pipelines/issues/755


advieser:
Noting for future reference, that R CMD check failed since `%check||%`/`%check&&%` could not be found when it was used inside of `custom_check` using `crate()`, i.e. `custom_check = crate(function(x) check_...(x) %check||% check_...(x)`.

mb706:
Apparently there was a bug in `crate()`, which is fixed in https://github.com/mlr-org/mlr3misc/pull/114 now. Until this is on cran, the necessary workaround is to specify the `.parent` argument of `crate()` directly. Could you see if using `crate(...., .parent = topenv())` works?
The reason we use `crate()` here is that we don't want the check functions to carry around the environment of the `initialize()`-call.

advieser:
A couple of parameters (`approx_pow`, `target_n_neighbors`, `target_weight`, `pca_method`) are only used if another parameter is NULL or non-NULL. The way I see it, this can't be tested for using the `depends` argument for the Domain, since if I use `depends = quote(approx_pow == NULL)`, I get 
```
Error in constructor(value) : 
  Assertion on 'rhs' failed: Must have length 1, but has length 0.
```
Obviously, `is.null(qpprox_pow)` can't be parsed either.
Did I get that correct?

There are other cases that can't be tested for using `depends`, however, this comes up often enough, that I thought I'd ask.

R/PipeOpUMAP.R
```diff
@@ -350,38 +362,69 @@ PipeOpUMAP = R6Class("PipeOpUMAP",
           tags = c("train", "umap"),
           custom_check = crate(function(x) check_list(x, types = c("integer", "numeric", "character"),
                                                       min.len = 1, max.len = 8, names = "unique", null.ok = TRUE))
+        ),
+        # Parameters that are passed to umap_transform to overwrite parameters from training for prediction
+        search_k_transform = p_int(default = NULL, special_vals = list(NULL), tags = c("predict", "overwrite"), depends = quote(nn_method == "annoy")),
+        n_epochs_transform = p_int(lower = 1L, default = NULL, special_vals = list(NULL), tags = c("predict", "overwrite")),
+        init_transform = p_fct(levels = c("weighted", "average"), special_vals = list("custom"), default = "weighted", tags = c("predict", "overwrite")),
+        init_transform_custom = p_uty(custom_check = check_matrix, tags = "predict", depends = quote(init_transform == "custom")),
```
mb706, line 370: this one has no tag 'overwrite' on purpose?

PR #795, opened 2024-08-14
Test reworked weights

mllg:
https://github.com/mlr-org/mlr3/pull/1065


mb706:
Is https://github.com/mlr-org/mlr3/pull/1065 done now? I see there are unanswered questions there...

Issue #790, opened 2024-08-06
Themis package functions


mb706:
what other methods exist in the themis package that we could use?

advieser:
`themis` implements differrent methods for handling unbalanced data. This is a summary based on the [manual](https://cran.r-project.org/web/packages/themis/themis.pdf):
- `adasyn` (Adaptive Synthetic Algorithm) which generates synthetic positive instances; works for tasks with only numeric features that have no `NA`s.
- `bsmote` (Borderline SMOTE) which generates new examples of the minority class using nearest neighbors of these
cases in the border region between classes; works for tasks with only numeric features that have no `NA`s.
- `nearmiss` which eliminates entries from the majority class that have the smallest distance to the minority class; works for tasks with only numeric features that have no `NA`s.
- `smote`, see `PipeOpSmote`
- `smotenc`, currently under development as `PipeOpSmoteNC`
- `tomek` which removes observations that are part of tomek links; works for tasks with only numeric features that have no `NA`s.

Furthermore, the package implements a recipe for one external function:
- `ROSE::ROSE()` which creates a sample of synthetic data by enlarging the features space of minority and majority class
examples; for inbalanced data in binary classification tasks.

Lastly, the package implements a recipe (`step_upsampling`) without a separate implementation, which will replicate rows of a data set to make the occurrence of levels in a specific factor level equal. Only intended for training.

advieser:
Of these, the following are also implemented in ``smotefamily`:
- `adasyn`
- `blsmote` (Borderline-Smote)
and additionally:
- `ANS` (Adaptive Neighbor SMOTE)
- `DBSMOTE` (Density-based SMOTE)
- `RSLS` (Relocating Safe-Level SMOTE)
- `SLS` (Safe-level SMOTE)

mb706:
We have the themis content itself already, so I will close this for now.

PR #817, opened 2024-08-31
New Down-Sampling PipoOps (Tomek, Nearmiss) based on `themis`

advieser:
This implements two new PipeOps for down-sampling inbalanced data, by calling `themis` functions:
- `PipeOpTomek`: Removes Tomek Links, i.e. pairs of observations that are nearest neighbors and of different classes. Note that this is only one possible implementation of Tomek Links, which is used for data cleaning. There also exists an algorithm for balancing data, in which not both observations of a Tomek Link are removed but only the majroity class member of the pair. However, this is the only version currently implemented by `themis`. 
- `PipeOpNearmiss`: Removes instances of the non-minority classes based on the NEARMISS algorithm, i.e. the instances that have the smallest mean distance to the closest instances of other classes. This is, again, only one possible implementation, but the only one in `themis`.
The documentation in `themis` seems to contain a few errors (probably due to being copied from another function).

As of right now, these two pipeops ignore stratification completely. 

I'm looking for some feedback about the way I filter the task based on the `themis` result. As the result is a `data.table`, called `dt`, I currently take the rownames of that result:
```r
keep = as.integer(row.names(dt))
task$filter(keep)
```
This seems a bit clunky to me. An alternative I thought of would be
```r
keep = as.integer(row.names(fintersect(task$data(), dt)))
```
which I'd expect to be more robust but also computationally more intensive (don't know how efficient `fintersect` is).

Partially addresses https://github.com/mlr-org/mlr3pipelines/issues/790.


R/PipeOpNearmiss.R
```diff
@@ -0,0 +1,109 @@
...
+      # Only numeric and integer features allowed
+      if (!all(task$feature_types$type %in% c("numeric", "integer"))) {
+        stop("Nearmiss does only accept numeric and integer features. Use PipeOpSelect to select the appropriate features.")
+      }
+
+      # Down-sample data
+      dt = setDT(invoke(themis::nearmiss, df = task$data(), var = task$target_names,
+                        .args = self$param_set$get_values(tags = "nearmiss")))
+
+      keep = as.integer(row.names(dt))
```
mb706, line 103:
```suggestion
      keep = task$row_ids[as.integer(row.names(dt))]
```

mb706: don't forget to test this!

R/PipeOpTomek.R
```diff
@@ -0,0 +1,98 @@
...
+      }
+      # Only numeric and integer features allowed
+      if (!all(task$feature_types$type %in% c("numeric", "integer"))) {
+        stop("Tomek does only accept numeric and integer features. Use PipeOpSelect to select the appropriate features.")
+      }
+
+      # Down-sample data
+      dt = setDT(invoke(themis::tomek, df = task$data(), var = task$target_names))
+
+      keep = as.integer(row.names(dt))
```
mb706, line 92: see above

R/PipeOpNearmiss.R
```diff
@@ -0,0 +1,109 @@
...
+
+    .train_task = function(task) {
+      cols = task$feature_names
+
+      # Return task unchanged, if no feature columns exist
+      if (!length(cols)) {
+        return(task)
+      }
+      # Only numeric and integer features allowed
+      if (!all(task$feature_types$type %in% c("numeric", "integer"))) {
```
mb706, line 95: only give error if there are no feature types of this kind; ignore non-numeric/integer cols instead

R/PipeOpNearmiss.R
```diff
@@ -0,0 +1,109 @@
...
+        stop("Nearmiss does only accept numeric and integer features. Use PipeOpSelect to select the appropriate features.")
+      }
+
+      # Down-sample data
+      dt = setDT(invoke(themis::nearmiss, df = task$data(), var = task$target_names,
```
mb706, line 100: `task$data(cols = task$feature_types[type %in% c("integer", "numeric"), id])`

R/PipeOpTomek.R
```diff
@@ -0,0 +1,98 @@
...
+
+    .train_task = function(task) {
+      cols = task$feature_names
+
+      # Return task unchanged, if no feature columns exist
+      if (!length(cols)) {
+        return(task)
+      }
+      # Only numeric and integer features allowed
+      if (!all(task$feature_types$type %in% c("numeric", "integer"))) {
```
mb706, line 85: see above

R/PipeOpTomek.R
```diff
@@ -0,0 +1,98 @@
...
+        stop("Tomek does only accept numeric and integer features. Use PipeOpSelect to select the appropriate features.")
+      }
+
+      # Down-sample data
+      dt = setDT(invoke(themis::tomek, df = task$data(), var = task$target_names))
```
mb706, line 90: see above

tests/testthat/test_pipeop_nearmiss.R
```diff
@@ -0,0 +1,49 @@
...
+  # Empty task is returned unchanged
+  task$select(character(0))
+  expect_equal(
+    op$train(list(task))[[1L]],
+    task
+  )
+
+  # PipeOp does not accept tasks with wrong feature types
+  task = tsk("german_credit")
+  expect_error(op$train(list(task)))
```
mb706, line 47: test that this is accepted

mb706: check that result is the same with factor cols removed?

R/PipeOpNearmiss.R
```diff
@@ -0,0 +1,109 @@
...
+      # Return task unchanged, if no feature columns exist
+      if (!length(task$feature_names)) {
+        return(task)
+      }
+      # At least one numeric or integer feature required
+      if (!any(task$feature_types$type %in% c("numeric", "integer"))) {
+        stop("Nearmiss needs at least one numeric or integer feature to work.")
+      }
+      # Subset columns to only include integer/numeric features and the target
+      cols = c(task$feature_types[get("type") %in% c("integer", "numeric"), get("id")], task$target_names)
```
mb706, line 98:
```suggestion
      type = id = NULL
      cols = c(task$feature_types[type %in% c("integer", "numeric"), id], task$target_names)
```

R/PipeOpTomek.R
```diff
@@ -0,0 +1,98 @@
...
+      # Return task unchanged, if no feature columns exist
+      if (!length(task$feature_names)) {
+        return(task)
+      }
+      # At least one numeric or integer feature required
+      if (!any(task$feature_types$type %in% c("numeric", "integer"))) {
+        stop("Tomek needs at least one numeric or integer feature to work.")
+      }
+      # Subset columns to only include integer/numeric features and the target
+      cols = c(task$feature_types[get("type") %in% c("integer", "numeric"), get("id")], task$target_names)
```
mb706, line 88:
```suggestion
      type = id = NULL
      cols = c(task$feature_types[type %in% c("integer", "numeric"), id], task$target_names)
```

PR #824, opened 2024-09-14
New Balancing PipeOps based on `smotefamily`

advieser:
Implements two new pipeops based on `smotefamily`:
- `PipeOpADAS`: ADASYN algorithm for balancing data. New instances are created based on difficulty for learning. 
- `PipeOpBLSmote`: Borderline-SMOTE algorithm for balancing data. New instances are created only for data ponts near the decision regions.

Both PipeOps only work for classification tasks with numeric features that have no missing values and accept target columns with multiple classes (although only the class with the least instances is upsampled).

Both pipeops use the implementation from https://github.com/mlr-org/mlr3pipelines/pull/815 for handling unsupported column types (so that has to be checked first).

Currently, `PipeOpBLSmote` prints `[1] "Borderline-SMOTE done"` after training. I'd think that we'd want to suppress this somehow, ideally?

Both `smotefamily::BLSMOTE` and `smotefamily::ADAS` contain bugs as of right now, which should be fixed on the side of `smotefamily`. This includes a bug where tasks with one feature cannot be handled (due to subsetting of the `data.frame` reducing the data to a vector on which `nrow()` is called which is then used in an if-statement).

Partially addresses https://github.com/mlr-org/mlr3pipelines/issues/790


R/PipeOpADAS.R
```diff
@@ -0,0 +1,115 @@
...
+      }
+      dt = task$data(cols = cols)
+
+      # calculate synthetic data
+      st = setDT(invoke(smotefamily::ADAS, X = dt, target = task$truth(),
+        .args = self$param_set$get_values(tags = "adas"),
+        .opts = list(warnPartialMatchArgs = FALSE))$syn_data)  # ADAS uses partial arg matching internally
+
+      # rename target column and fix character conversion
+      st[["class"]] = as_factor(st[["class"]], levels = task$class_names)
```
mb706, line 107: this can create duplicate column 'class'. Instead rely on the fact that the new class is the last column of st. Test that this works with second 'class'-column (ideally not a factor column, so the 'as_factor()' would be noticed). 

Should also be checked in the other smotefamily functions

R/PipeOpBLSmote.R
```diff
@@ -0,0 +1,124 @@
...
+  inherit = PipeOpTaskPreproc,
+  public = list(
+    initialize = function(id = "blsmote", param_vals = list()) {
+      ps = ps(
+        K = p_int(lower = 1, default = 5, tags = c("train", "blsmote")),
+        C = p_int(lower = 1, default = 5, tags = c("train", "blsmote")),
+        # dup_size = 0 leads to behaviour different from 1, 2, 3, ..., because it means "duplicating until balanced", so it is a 'special_vals'.
+        dupSize = p_int(lower = 1, default = 0, special_vals = list(0), tags = c("train", "blsmote")),
+        # Default of `method` is derived from the source code of smotefamily::BLSMOTE(), not documented there.
+        method = p_fct(levels = c("type1", "type2"), tags = c("train", "blsmote"))
```
mb706, line 86: set default

R/PipeOpBLSmote.R
```diff
@@ -0,0 +1,124 @@
...
+      dt = task$data(cols = cols)
+
+      # Calculate synthetic data
+      # TODO: Do we have a way to suppress messages by print()?
+      st = setDT(invoke(smotefamily::BLSMOTE, X = dt, target = task$truth(),
+        .args = self$param_set$get_values(tags = "blsmote"),
+        .opts = list(warnPartialMatchArgs = FALSE))$syn_data)  # BLSMOTE uses partial arg matching internally
+
+      # Rename target column and fix character conversion
+      st[["class"]] = as_factor(st[["class"]], levels = task$class_names)
```
mb706, line 116: see above

R/PipeOpBLSmote.R
```diff
@@ -0,0 +1,124 @@
...
+  public = list(
+    initialize = function(id = "blsmote", param_vals = list()) {
+      ps = ps(
+        K = p_int(lower = 1, default = 5, tags = c("train", "blsmote")),
+        C = p_int(lower = 1, default = 5, tags = c("train", "blsmote")),
+        # dup_size = 0 leads to behaviour different from 1, 2, 3, ..., because it means "duplicating until balanced", so it is a 'special_vals'.
+        dupSize = p_int(lower = 1, default = 0, special_vals = list(0), tags = c("train", "blsmote")),
+        # Default of `method` is derived from the source code of smotefamily::BLSMOTE(), not documented there.
+        method = p_fct(levels = c("type1", "type2"), tags = c("train", "blsmote"))
+      )
```
mb706, line 87: add hyperparameter "quiet", set to TRUE, *on default*, use `capture.output()` to suppress print in train.

Issue #723, opened 2023-05-22
filter hash not entirely correct


mb706:
depends on https://github.com/mlr-org/mlr3filters/issues/154

mb706:
use filter$hash here: https://github.com/mlr-org/mlr3pipelines/blob/f97409c0ce5e4330c0dd57f50b39d27b52e4e16d/R/PipeOpFilter.R#L198

but first need to fix filter hash in mlr3filters

PR #827, opened 2024-09-22
fix: hash for filters

advieser:
addresses #723 

depends on https://github.com/mlr-org/mlr3filters/pull/168


mb706:
waiting for filters cran release now

advieser:
TODO before merging: Check that version number for `mlr3filters` is correct for the next CRAN release (depending on whehter it is a minor or patch)

Issue #567, opened 2021-02-20
PipeOpSubsample: allow stratification by other columns, eg. group column


mb706:

mb706:
hyperparameter `use_groups`, initialized to `TRUE`? Would only change current behaviour if a group column is actually present.

PR #834, opened 2024-10-03
New param `use_groups` for `PipeOpSubsample` and rework for `task_filter_ex()`

advieser:
Closes https://github.com/mlr-org/mlr3pipelines/issues/567

If `use_groups = TRUE` (default), we subsample whole groups. This leads to `frac` not being fully accurate.
We currently don't support stratification (`stratify = TRUE`) and subsampling grouped data at the same time, same as with Resamplings in mlr3.
This changes the default behavior for tasks with a column with role "group".

Right now, `task$row_roles$use` is not respected when `use_groups = TRUE`. Question is, how we would want to handle that? If a group contains any row that is not in `task$row_roles$use` , we ignore the group for subsampling?


R/PipeOpSubsample.R
```diff
@@ -67,30 +83,38 @@ PipeOpSubsample = R6Class("PipeOpSubsample",
...
-      if (!self$param_set$values$stratify) {
-        keep = shuffle(task$row_roles$use,
-          ceiling(self$param_set$values$frac * task$nrow),
-          replace = self$param_set$values$replace)
-      } else {
+      pv = self$param_set$get_values(tags = "train")
+
+      if (pv$stratify && pv$use_groups) {
+        stop("Cannot combine stratification with grouping")
+      } else if (!pv$stratify && pv$use_groups && !is.null(task$groups)) {
```
mb706, line 100:
```suggestion
      } else if (pv$use_groups && !is.null(task$groups)) {
```

R/PipeOpSubsample.R
```diff
@@ -67,30 +83,38 @@ PipeOpSubsample = R6Class("PipeOpSubsample",
...
+        group = NULL  # binding for CHECK
+        grps = unique(task$groups$group)
+        # some way to use row_roles = use? Ignore group that has one row with row_roles = use
+        keep_grps = shuffle(grps, ceiling(pv$frac * length(grps)), replace = pv$replace)
+        keep = task$groups[group %in% keep_grps]$row_id
+      } else if (pv$stratify && !pv$use_groups) {
```
mb706, line 106:
```suggestion
      } else if (pv$stratify) {
```

R/PipeOpSubsample.R
```diff
@@ -67,30 +83,38 @@ PipeOpSubsample = R6Class("PipeOpSubsample",
...
-      } else {
+      pv = self$param_set$get_values(tags = "train")
+
+      if (pv$stratify && pv$use_groups) {
+        stop("Cannot combine stratification with grouping")
+      } else if (!pv$stratify && pv$use_groups && !is.null(task$groups)) {
+        group = NULL  # binding for CHECK
+        grps = unique(task$groups$group)
+        # some way to use row_roles = use? Ignore group that has one row with row_roles = use
+        keep_grps = shuffle(grps, ceiling(pv$frac * length(grps)), replace = pv$replace)
```
mb706, line 104: Lets do this properly:
* get a vector of group sizes (only respecting the rows that are in row_roles = use -- the ones that are not should be treatet like they are not present. Also try to handle the case where there are duplicate values in row_roles = use; treat them like samples that are present multiple times.
* for sampling without replacement: shuffle the order of groups, cut off of at the place where abs(cumsum(groupsizes[order]) / sum(groupsizes) - frac) is somehow minimal.
* for sampling with replacement: sample with `prob = 1 / groupsizes`, so that individual samples are sampled uniformly. maybe this has to happen in a loop until sum(groupsizes[sampled]) approaches frac * task$nrow?

tests/testthat/test_pipeop_subsample.R
```diff
@@ -73,7 +73,40 @@ test_that("PipeOpSubsample works stratified", {
...
+
+  op$param_set$set_values(frac = 2, replace = TRUE)
+  train_out = op$train(list(task))[[1L]]
+  # grouped data are kept together
+  grps_out = table(train_out$groups$group)
+  expect_equal(
+    grps_out,
+    table(task$groups$group)[names(grps_out)]
+  )
+
```
mb706, line 104: test that fraction is adhered to as well as possible; even with imbalanced group sizes (e.g. one very large group).

mb706 (dismissed):

R/PipeOpSubsample.R
mb706: > I'll probably have to change `task_filter_ex()` for this to work.

I think that makes sense. Maybe it should be generalized to work properly with all column roles (group, weight, ...) that don't explicitly require unique values.

R/PipeOpSubsample.R
```diff
@@ -97,13 +97,22 @@ PipeOpSubsample = R6Class("PipeOpSubsample",
...
+        # task$groups automatically removes rows not in task$row_roles$use and allows rows to be included multiple times
+        grp_sizes = table(task$groups$group)
+
+        # If we sample with replacement, drawing groups of larger sizes should be less probable
+        probs = if (pv$replace) 1 / grp_sizes else NULL
+
+        # We randomly shuffle the groups and keep all up to the group for
+        # which the fraction of rows is closest to the desired fraction.
+        shuffled = shuffle(grp_sizes, prob = probs, replace = pv$replace)
+        cutoff_index = which.min(abs(cumsum(shuffled) / sum(shuffled) - frac))
```
mb706, line 110: should this be `pv$frac`?

R/PipeOpSubsample.R
mb706: Tbh, I think we want to sample all groups uniformly even in the case with replacement. So I think the `prob` thing should just go...

mb706:
todo: test with task where trailing rows were `$filter()`ed out

R/utils.R
```diff
@@ -24,24 +24,64 @@ calculate_collimit = function(colwidths, outwidth) {
...
+        while (length(groups) < .N) {
+          new_group = paste0(group[[1]], "_", i)
+          if (new_group %nin% grps) groups[[length(groups) + 1]] = new_group
+          i = i + 1
+        }
+        groups
+      }, by = row_id]
+
+      # Use "new_groups" to update the group entries.
+      new_data[, (task$col_roles$group) := new_groups$group]
```
mb706, line 72: Test cases we want to check: Consider task with 3 rows, group 'a', 'b', 'c'; row roles "use" c(1, 1, 2, 2, 3)
* row_ids c(1, 1, 2, 2, 3) -> we get flattened task that has the same groups as before (has ids 1, 2, 3, 4, 5, groups 'a', 'b', 'c', 'a', 'b')
* row_ids c(1, 1, 1, 1, 2, 2, 3) -> add group a_1, i.e. task has c(1, 2, 3, 4, 5, 6, 7), groups 'a', 'b', 'c', 'a', 'b', 'a_1', 'a_1'
* row_ids c(1, 1, 1, 1, 3) -> same as above, but 'b' disappears: c(1, 2, 3, 4, 5), groups 'a', 'a', 'c', 'a_1', 'a_1'
* row_ids 3 -> groups a and b disappear: task has row-id 3, group 'c'

mb706: * row_ids c(1, 1, 1, 1, 3, 1, 1) -> c(1, 3, 4, 5, 6, 7, 8), groups 'a', 'c', 'a', 'a_1', 'a_1', 'a_2', 'a_2'

PR #838, opened 2024-10-09
PipeOpLearnerCVPlus

lona-k:


R/PipeOpLearnerCVPlus.R
```diff
@@ -0,0 +1,204 @@
...
+      pv = private$.cvplus_param_set$values
+
+      # Compute CV Predictions
+      rdesc = rsmp("cv", folds = pv$folds)
+      rr = resample(task, private$.learner, rdesc, store_models = TRUE)
+
+      prds = rbindlist(map(rr$predictions(predict_sets = "test"), as.data.table), idcol = "fold")
+
+      # Add trained models and residuals to PipeOp state
+      self$state = list(cv_models = rr$learners,
```
mb706, line 164:
```suggestion
      self$state = list(cv_model_states = map(rr$learners, "state"),
```

mb706: It is best to only store the essential information in the PipeOp's `$state`, which are the model states. Try not to store the entire learner objects.

R/PipeOpLearnerCVPlus.R
```diff
@@ -0,0 +1,204 @@
...
+    learner_model = function(val) {
+      if (!missing(val)) {
+        if (!identical(val, private$.learner)) {
+          stop("$learner_model is read-only.")
+        }
+      }
+      if (is.null(self$state) || is_noop(self$state)) {
+        private$.learner
+      } else {
+        multiplicity_recurse(self$state, clone_with_state, learner = private$.learner)
```
mb706, line 141: this does not work, since the PipeOp's `$state` is not the same as the Learner's `$state`; this is different in the PipeOpLearnerCV. But tbh I don't know what kind of result one would want here, maybe best to discuss this.

One possibility of what could work would be
```r
multiplicity_recurse(self$state, learner = function(state) {
  map(state$cv_model_states, clone_with_state, learner = private$.learner)
})
```

R/PipeOpLearnerCVPlus.R
```diff
@@ -0,0 +1,204 @@
...
+                        residuals = prds[, .(fold, residual = abs(truth - response))])
+
+      list(NULL)
+    },
+    .predict = function(task) {
+      pv = private$.cvplus_param_set$values
+
+      mu_hat = map(self$state$cv_models, function(learner) {
+          as.data.table(learner$predict(task))
+        })
```
mb706, line 174:
```suggestion
      mu_hat = map(self$state$cv_model_states, function(state) {
          on.exit({private$.learner$state = NULL})
          private$.learner$state = state
          as.data.table(private$.learner$predict(task))
        })
```

tests/testthat/test_pipeop_learnercvplus.R
```diff
@@ -0,0 +1,92 @@
...
+  expect_equal(dim(predictions$quantiles), c(task$nrow, 3))  # lower and upper quantiles
+
+  expect_error(PipeOpLearnerCVPlus$new())
+
+  expect_pipeop(po("learner_cv_plus", lrn("regr.featureless")))
+})
+
+test_that("PipeOpLearnerCVPlus - param set and values", {
+  skip_if_not_installed("ranger")
+  lrn = mlr_learners$get("regr.ranger")
```
mb706, line 31: do you need 'ranger' for these tests? In that case, you need to load `mlr3learners` (and don't forget to `skip_if_not_installed` it), using `loadNamespace()` or `requireNamespace()`, otherwise the "ranger"-learner is not added to the `lrn()` dictionary.

mb706: If you can use `rpart` instead of `ranger` then this would not be necessary, because rpart is in the `lrn()` dictionary from the start, without `mlr3learners`.

R/PipeOpLearnerCVPlus.R
mb706: You'd also need to update the documentation.

mb706 (changes requested):
I think the reasons why the tests fail is that `lrn("regr.ranger")` is not in the `lrn()` dictionary when `mlr3learners` is not loaded. So you either have to load that, or see if you can use `lrn("regr.rpart")` instead, which is present when using mlr3 without mlr3learners.

mb706:
Also don't forget to add a NEWS entry. And feel free to add yourself as `ctb` in the `DESCRIPTION` if you want.

R/PipeOpLearnerCVPlus.R
```diff
@@ -0,0 +1,204 @@
...
+#' \dontshow{ \} }
+PipeOpLearnerCVPlus = R6Class("PipeOpLearnerCVPlus",
+  inherit = PipeOp,
+  public = list(
+    initialize = function(learner, id = NULL, param_vals = list()) {
+      private$.learner = as_learner(learner, clone = TRUE)
+      id = id %??% private$.learner$id
+      type = private$.learner$task_type
+
+      if (!("regr" %in% type)) {
```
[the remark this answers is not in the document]
lona-k: No, I just wasn't aware that `task_type` is always `character(1)`, I changed it

R/PipeOpLearnerCVPlus.R
lona-k: Okay, I changed the state and documentation to only include the state.

tests/testthat/test_pipeop_learnercvplus.R
lona-k: I don't need `ranger`, I updated it to use `rpart` in all tests.

R/PipeOpLearnerCVPlus.R
```diff
@@ -0,0 +1,204 @@
...
+        stop("PipeOpLearnerCVPlus only supports regression.")
+      }
+
+      task_type = mlr_reflections$task_types[type, mult = "first"]$task
+      out_type = mlr_reflections$task_types[type, mult = "first"]$task
```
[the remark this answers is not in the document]
lona-k: `out_type` was necessary here, but I changed it to `mlr_reflections$task_types[type, mult = "first"]$prediction` like you suggested.

Issue #835, opened 2024-10-03
PipeOpDecode


mb706:
Inverses one-hot-encoding: creates a factor-column that indicates which of the input numeric columns has the maximal value.

Should have argument 'treatment_encoding' (init: FALSE): if TRUE, it includes an additional level if all cols are 0 (becoming the inverse of PipeOpEncode with method == treatment).

Should also have an argument `group_pattern`, a regular expression. The `group_pattern` is applied to all col names and the first regex group is extracted. All columns that have different value here are treated separately from each other. The levels that are created then correspond to `gsub(group_pattern, "", colnames())`. Initialized as `"^([^.]*)\\."`.

The point here is that we may have columns `x.a`, `x.b`, `x.c`, `y.a`, `y.b`. The  `"^([^.]*)\\."`-match matches `"x"` for the first three cols and creates levels `a`, `b`, and `c`. It then matches `"y"` for the last two cols, creating the factor cols with levels `a` and `b`. Should the user e.g. have columns `x_a`, `x_b`, ..., then this would need to be changed to `"^([^_]*)_"`. Should the user not want any groups, and instead get levels `x.a`, `x.b`, ..., `y.b` in a single result column, the pattern would be `""`.

If the pattern is not `""`, we ignore all columns that do not match the `group_pattern`; I am assuming that this is what a user wants basically all of the time, even though it unfortunately undermines the `affect_columns` argument somewhat.

mb706:
suggestion for `state`:

* named list, named by columns that are being created, with content for each such column:
  * named list, named by the name of input columns, containing `character(1)` in each entry, naming the name of the resulting factor.

also maybe the content of the `treatment_encoding` flag for prediction, since changing the hyperparamter after training is not allowed to have an effect.

probably good idea to use PipeOpTaskPreprocSimple.

mb706:
in the `x.a`, `x.b`, `x.c`, `y.a`, `y.b`, the state would be
```r
list(
  colmaps = list(
    x = c(x.a = "a", x.b = "b", x.c = "c"),
    y = c(y.a = "a", y.b = "b")
  ),
  treatment_encoding = FALSE
)
```

mb706:
maybe `treatment_encoding` flag is not necessary, since we can see this from the fact that there are entries in the col-maps with empty name.

PR #839, opened 2024-10-21
New PipeOpDecode to reverse encoding

advieser:
Adds a new `PipeOpDecode` to reverse one-hot and treatment encoding.

Closes https://github.com/mlr-org/mlr3pipelines/issues/835


R/PipeOpDecode.R
```diff
@@ -0,0 +1,240 @@
...
+
+    .get_state_dt = function(dt, levels, target) {
+      pv = self$param_set$values
+      cols = colnames(dt)
+
+      if (pv$treatment_encoding) {
+        # Determine name for reference level
+        ref_name = "ref"
+        counter = 1
+        while (ref_name %in% cols) {
```
mb706, line 136: This should be a function that is then called at each occurrence of `... = ref_name`, since the collision check needs to be done for each individual group separately.

R/PipeOpDecode.R
```diff
@@ -0,0 +1,240 @@
...
+        }
+        # Replace occurrences of old column names with corresponding new level names
+        set(dt, , new_col, factor(lvls[match(dt[[new_col]], names(lvls))], levels = lvls))
+      }
+
+      # Drop old columns
+      drop = unlist(lapply(colmaps, names))
+      # If existent, remove empty string elements
+      drop = discard(drop, drop == "")
+      dt[, (drop) := NULL]
```
mb706, line 233: add new cols after deleting old ones, in case there is name overlap

Issue #813, opened 2024-08-26
regression tree based binning of cts features


mb706:
https://github.com/yandex-research/rtdl-num-embeddings/blob/abf8a8b35854e4b06476bb48902096b0b58ffce2/package/rtdl_num_embeddings.py#L300

https://arxiv.org/abs/2203.05556#

Could either be its own pipeop or an extension of an existing binning pipeop

PR #861, opened 2024-12-20
New `PipeOpEncodePL` for Piecewise Linear Encoding

advieser:
closes https://github.com/mlr-org/mlr3pipelines/issues/813


R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,151 @@
...
+    initialize = function(task_type, id = "encodepl", param_vals = list()) {
+      # NOTE: Might use different name, change assert, and conditions
+      assert_choice(task_type, mlr_reflections$task_types$task)
+      if (task_type == "TaskRegr") {
+        private$.tree_learner = LearnerRegrRpart$new()
+      } else if (task_type == "TaskClassif") {
+        private$.tree_learner = LearnerClassifRpart$new()
+      } else {
+        stopf("Task type %s not supported", task_type)
+      }
```
advieser, line 64: Don‘t need `task_type` if we use `method = "quantiles"`, meaning that we should only check this in that case and don‘t throw an error, etc. We could set task type as `"Task"` by default …
However, `mlr_reflections$task_types` does not contain this as an acceptable name, so change the assert.

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,151 @@
...
+        })
+      } else {
+        learner = private$.tree_learner
+
+        bins = list()
+        for (col in cols) {
+          t = task$clone(deep = TRUE)$select(col)
+          splits = learner$train(t)$model$splits
+          # Get column "index" in model splits
+          boundaries = unname(sort(splits[, "index"]))
```
advieser, line 104: Can simplify this into one line now

advieser:
Consider whether we need to specify `.additional_phash_input` in tree pipeop.

advieser:
Fix column naming. Currently, column names are duplicated, with this behavior being introduced through the helper function.

advieser:
> Fix column naming. Currently, column names are duplicated, with this behavior being introduced through the helper function.

`imap()` sets names. Also try out `imap_dtc()`.

advieser:
Do we want to add a public method to retrieve the learner?

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+  private = list(
+
+    .get_bins = function(task, cols) {
+      stop("Abstract.")
+    },
+
+    .get_state = function(task) {
+      cols = private$.select_cols(task)
+      if (!length(cols)) {
+        return(list(bins = numeric(0)))  # early exit
```
mb706, line 96:
```suggestion
        return(list(bins = named_list()))  # early exit
```

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+#' # Calculated bin boundaries per feature
+#' pop$state$bins
+#' # Each feature was split into three encoded features using piecewise linear encoding
+#' train_out$head()
+PipeOpEncodePLQuantiles = R6Class("PipeOpEncodePLQuantiles",
+  inherit = PipeOpEncodePL,
+  public = list(
+    initialize = function(id = "encodeplquantiles", param_vals = list()) {
+      ps = ps(
+        numsplits = p_int(lower = 2, default = 2, tags = c("train", "predict")),
```
mb706, line 228:
```suggestion
        numsplits = p_int(lower = 2, init = 2, tags = c("train", "predict", "required")),
```

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+        type = p_int(lower = 1, upper = 9, default = 7, tags = c("train", "predict"))
+      )
+      super$initialize(id, param_set = ps, param_vals = param_vals, packages = "stats")
+    }
+  ),
+  private = list(
+
+    .get_bins = function(task, cols) {
+      numsplits = self$param_set$values$numsplits %??% 2
```
mb706, line 237:
```suggestion
      numsplits = self$param_set$values$numsplits
```

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+      # Defaulting to default value in stats::quantile, i.e. method 7
+      type = self$param_set$values$type %??% 7
+
+      lapply(task$data(cols = cols), function(d) {
+        unique(c(
+          min(d, na.rm = TRUE),
+          stats::quantile(d, seq(1, numsplits - 1) / numsplits, na.rm = TRUE, names = FALSE, type = type),
+          max(d, na.rm = TRUE)
```
mb706, line 245:
```suggestion
        stats::quantile(d, seq(0, 1, length.out = numsplits + 1), na.rm = TRUE, names = FALSE, type = type),
```

please make tests that this works for all the different 'type's

mb706: in particular that it includes mins and max

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+
+#' @title Piecewise Linear Encoding using Decision Trees
+#'
+#' @usage NULL
+#' @name mlr_pipeops_encodepltree
+#' @format [`R6Class`][R6::R6Class] object inheriting from [`PipeOpEncodePL`] / [`PipeOpTaskPreprocSimple`]/[`PipeOpTaskPreproc`]/[`PipeOp`].
+#'
+#' @description
+#' Encodes `numeric` and `integer` feature columns using piecewise lienar encoding. For details, see documentation of
+#' `PipeOpEncodePL` or the paper referenced below.
```
mb706, line 262: can you put a reference here ("see x et al.")?

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+      learner = private$.tree_learner
+      bins = list()
+      for (col in cols) {
+        t = task$clone(deep = TRUE)$select(col)
+        model = learner$train(t)$model
+        frame = model$frame
+
+        # Extracting splits according to https://github.com/cran/rpart/blob/983544575b80df286bf8ce238faf4afa145872cd/R/labels.rpart.R#L26
+        # NOTE: Maybe this isn't necessary, and we can just take index of splits (should splits exist), since compete and
+        # surrogate *might* only be larger than 0 than trained with more than one feature?
```
mb706, line 381:
```suggestion
        # NOTE: `frame$nsurrogate` and `frame$ncompete` seem to be always 0 in our case, since competing and
        # surrogate splits only exist when there are more than one freature. Leaving it in for completeness.
```

R/PipeOpEncodePL.R
```diff
@@ -0,0 +1,409 @@
...
+        # - surrogates could only be used, if multiple variables exist (cannot use a non-existing variable as surrogate
+        #   if the first considered variables has NAs)
+        # - competes could not occur with one feature, since no alternative splits are possible
+        if (nrow(frame) > 1L) {  # do splits exist?
+          is_leaf = frame$var == "<leaf>"
+          frame = frame[!is_leaf, ]
```
mb706, line 387:
```suggestion
          frame = frame[!is_leaf, , drop = FALSE]
```

tests/testthat/test_pipeop_encodepl.R
```diff
@@ -0,0 +1,206 @@
...
+  # Test default parameters
+  train_out = op$train(list(task))[[1L]]
+  predict_out = op$predict(list(task))[[1L]]
+
+  # bins are correct
+  expect_equal(op$state$bins, list(x = c(2, 7, 12), y = c(10, 50, 95)))
+
+  # encoding and column naming
+  dt_encoded = data.table(
+    x.bin1 = c(0, 1, 2/5, 1, 4/5, 1, NA, NA),
```
mb706, line 26: spaces between operators

R/PipeOpTrafo.R
```diff
@@ -233,7 +233,7 @@ PipeOpTargetInvert = R6Class("PipeOpTargetInvert",
     initialize = function(id = "targetinvert", param_vals = list()) {
       super$initialize(id = id, param_vals = param_vals,
         input = data.table(name = c("fun", "prediction"), train = c("NULL", "NULL"), predict = c("function", "Prediction")),
-        output = data.table(name = "output", train = "NULL", predict = "Prediction")
+        output = data.table(name = "output", train = "NULL", predict = "Prediction"), tags = "target transformation"
```
mb706, line 236:
```suggestion
        output = data.table(name = "output", train = "NULL", predict = "Prediction"), tags = "target transform"
```

Issue #872, opened 2025-01-28
Clean up PipeOp documentation


mb706:
the documentation of @methods is a bit unclean.

split this up: should still only be one "Methods" section, but have some non-bulletpoint text saying to implement a pipeop, the following internal private methods need to be implemented, followed by bullet point list describing private methods in the usual way.

Don't duplicate too much wiht internals, but some duplication is okay.

Issue #871, opened 2025-01-28
document columns of as.data.table(po())


mb706:
* document what the columns mean
* short documentation of the types in the type columns
  * 'NULL', null object, but *means* no data, e.g. output of learner during train
  * '*' means any type, e.g. pipeopnop, pipeopcopy
  * `[ ]` means multiplicity, link to multiplicity
* input number columns: document 'NA' means number depends on construction argument.

mb706:
(also document other columns I dont mention here obvs)

Issue #868, opened 2025-01-22
Miscellaneous documentation corrections


advieser:
`PipeOpTaskPreprocSimple` has the same construction as `PipeOpTaskPreproc`. However, `tags` and `feature_types` are only documented for `PipeOpTaskPreproc`.

advieser:
Add backticks for some Fields of `PipeOp`

PR #873, opened 2025-02-06
Various docs changes

advieser:
Docs standardization (closes https://github.com/mlr-org/mlr3pipelines/issues/868):
- Add `Fields` section where it was missing (mostly in PipeOps inheriting from `PipeOpTaskPreprocSimple`)
- Reference PipeOps from which we inherit in docs as follows:
 	- `@format`: list all
 	- Otherwise, only list those that changed / added something for the respective section.
 	 	- for PipeOps inheriting from `PipeOpTaskPreproc` or `PipeOpTaskPreprocSimple` this usually means to reference `PipeOpTaskPreproc` except in `@section Methods` where we would usually write ```[`PipeOpTaskPreprocSimple`]/[`PipeOpTaskPreproc`]/[`PipeOp`]``` or ```[`PipeOpTaskPreproc`]/[`PipeOp`]```, respectively.
 - Removed self-referencing links (e.g. ```[`PipeOp`]``` in docs for `PipeOp`)
 
 Cleaned up docs of methods in PipeOp (closes https://github.com/mlr-org/mlr3pipelines/issues/872) and added documentation for field `label`.
 
Documented columns of `mlr_pipeops` (closes https://github.com/mlr-org/mlr3pipelines/issues/871).


advieser:
@mb706 Any idea what this should've become?
https://github.com/mlr-org/mlr3pipelines/blob/3485877fdab25089847f96afa4fc872502109a6a/R/PipeOpClassifAvg.R#L23

mb706:
But in seriousness, I think it is nothing and can probably be deleted...

mb706:
At least I can't think of anything that would fit in that context. Maybe I started the 'If `TRUE`' sentence from [this commit](https://github.com/mlr-org/mlr3pipelines/commit/8e21fbb096b757d671afc9685dd215f282b9f1d3) and changed my mind about where it should go and forgot to delete it...

advieser:
Probably, sounds plausible. Thanks!

mb706:
to do:

* no self-links in pipeop doc
* news entry

advieser:
We have a couple of PipeOps without the `@name` specification. These can be easily identified by looking into `man/`. Is this intended (as these seem to be the abstract classes). If so, should `PipeOpEncodePL` also have this?

mb706:
For future reference: `mlr_pipeops_<id>` only for PipeOps that can be constructed using `po("<id>")`, the others (i.e. abstract PipeOps) should just have `PipeOpXxx` (no special name).

advieser:
TODO:
- rename EncodePL
- add abstract pos in pkgdown

advieser:
Fixed the parsing of `PipeOpNMF`'s examples. This was caused by using `#` in a `\dontshow{}` line. However, when building the website locally, the code in the `\dontshow` line gives a warning which gets printed on the website. Should we suppress warnings for this, or is there a better alternative?

PR #874, opened 2025-02-07
feat: hotstart

be-marc:
Nothing to merge yet but a poc we should discuss. `mlr3tuning` is ready to use hotstarting efficiently and @sebffischer wants to use hotstarting for torch. This currently only works for pipelines that are deterministic until the hotstart model is reached. 

Further thoughts on this. Actually, the task would no longer have to be sent through the pipeline during hotstarting. If you cache the task before the hotstart model, you can simply continue there. However, this would somehow require the task to be cached at this point during the first training.

### DIFF
--- a/R/Graph.R
+++ b/R/Graph.R
@@ -449,10 +449,12 @@ Graph = R6Class("Graph",
       self$set_names(ids, sprintf("%s%s%s", assert_string(prefix), ids, assert_string(postfix)))
       invisible(self)
     },
+
     train = function(input, single_input = TRUE) {
       graph_load_namespaces(self, "train")
       graph_reduce(self, input, "train", single_input)
     },
+
     predict = function(input, single_input = TRUE) {
       graph_load_namespaces(self, "predict")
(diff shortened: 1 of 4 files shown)


be-marc:
```r
library(mlr3pipelines)
library(mlr3tuning)
library(mlr3learners)

task = tsk("german_credit")
resampling = rsmp("cv", folds = 3)
resampling$instantiate(task)

learner = as_learner(po("encodeimpact") %>>% lrn("classif.xgboost", nrounds = 3))

rr = resample(task, learner, resampling, store_models = TRUE)
rr$score(msr("classif.acc"))
rr$learners[[1]]$model$classif.xgboost

learner_2 = as_learner(po("encodeimpact") %>>% lrn("classif.xgboost", nrounds = 27))
learner_2$hotstart_stack = HotstartStack$new(rr$learners)

rr = resample(task, learner_2, resampling, store_models = TRUE, allow_hotstart = TRUE)
rr$score(msr("classif.acc"))
rr$learners[[1]]$model$classif.xgboost
``

Issue #631, opened 2021-11-20
PipeOpTargetTrafo drops missing factor levels


be-marc:
`PipeOpTargetTrafo` drops missing factor levels in task, whereas `mlr3` keeps the factor levels.

```r
library(mlr3)
library(mlr3pipelines)
options(mlr3.debug = TRUE)

task = tsk("boston_housing")
learner = lrn("regr.rpart")
ppl = ppl("targettrafo",
  graph = learner,
  targetmutate.trafo = function(x) log(x),
  targetmutate.inverter = function(x) list(response =  expm1(x$response)))
graph_learner = as_learner(ppl)

# fails
resample(task, graph_learner, rsmp("holdout"))

# > Error in model.frame.default(Terms, newdata, na.action = na.action, xlev = attr(object,  : 
# >  factor town has new levels Dover, Duxbury, Hamilton, Manchester, Marshfield, Medfield, Millis, Nahant, Weston
# > This happened PipeOp regr.rpart's $predict()

# works
resample(task, learner, rsmp("holdout"))
```

Using `PipeOpFixFactors` fixes the issue but maybe `PipeOpTargetTrafo`  should not drop the factor levels?

The gallery post bike sharing fails (https://github.com/mlr-org/mlr3gallery/issues/119).

be-marc:
`PipeOpFixFactors` seems not to  fix issue in the gallery post but in the above example.

sumny:
I think this could either be fixed in `mlr3pipelines` "manually" or we directly fix `mlr3::convert_task` which seems to cause the problem (i.e., during the trafo a new `Task` is created using the `DataBackend` of the task but during resampling it can happen that some levels are no longer present and therefore are also no longer present in the backend resulting in also being missing in the new `Task`.

Issue #707, opened 2023-01-24
PipeOpTargetMutate QOL


mb706:
- don't assert on function with only one argument, we can't set `trafo = log` currently.
- give the head of the task to allow conversion between tasks.

mb706:
overall: maybe play around with target mutate and discover the rough edges

advieser:
Point 1 was addressed by https://github.com/mlr-org/mlr3pipelines/pull/876.

PR #876, opened 2025-02-15
fix: Add `drop_levels = FALSE` to call of `mlr3::convert_task()` in `PipeOpTargetMutate` and `PipeOpTargetTrafoScaleRange`

advieser:
* Fixes dropping of unseen factor levels in `PipeOpTargetMutate` and `PipeOpTargetTrafoScaleRange`.
* Addresses point 1 from https://github.com/mlr-org/mlr3pipelines/issues/707.
* Checks that `trafo` hyperparameter of `PipeOpTargetMutate` returns a df/dt/matrix.

closes https://github.com/mlr-org/mlr3pipelines/issues/631


advieser:
We might instead want to add `drop_levels` as a parameter?

advieser:
Also, I don't know whether we want to add the [bikesharing example](https://github.com/mlr-org/mlr3gallery/tree/main/_inst/2020-07-27-bikesharing-demand) to mlr3website, now that it should work again? It was removed from mlr3gallery in 2021 because of this issue (see [here](https://github.com/mlr-org/mlr3gallery/issues/119)).

PR #880, opened 2025-02-20
mean -> median for docs describing stats::mad

### LINKED ISSUE #879: PipeOpScale docs say robust scaling uses mean absolute deviation
The sentence in question: "If robust is TRUE, this is the mean absolute deviation multiplied by 1.4826 (see [stats::mad](https://rdrr.io/r/stats/mad.html) of each feature, possibly after centering."

I think this should be the median absolute deviation, based on the [`stats::mad()` docs](https://rdrr.io/r/stats/mad.html). This would also be consistent with the description in the "Parameters" section.

cxzhang4:
resolves #879

### DIFF
--- a/R/PipeOpScale.R
+++ b/R/PipeOpScale.R
@@ -32,7 +32,7 @@
 #' * `scale` :: `numeric`\cr
 #'   The value by which features are divided. 1 if `scale` is `FALSE`\cr
 #'   If `robust` is `FALSE`, this is the root mean square, defined as `sqrt(sum(x^2)/(length(x)-1))`, of each feature, possibly after centering.
-#'   If `robust` is `TRUE`, this is the mean absolute deviation multiplied by 1.4826 (see [stats::mad] of each feature, possibly after centering.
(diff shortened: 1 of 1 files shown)


R/PipeOpScale.R
```diff
@@ -32,7 +32,7 @@
 #' * `scale` :: `numeric`\cr
 #'   The value by which features are divided. 1 if `scale` is `FALSE`\cr
 #'   If `robust` is `FALSE`, this is the root mean square, defined as `sqrt(sum(x^2)/(length(x)-1))`, of each feature, possibly after centering.
-#'   If `robust` is `TRUE`, this is the mean absolute deviation multiplied by 1.4826 (see [stats::mad] of each feature, possibly after centering.
+#'   If `robust` is `TRUE`, this is the median absolute deviation multiplied by 1.4826 (see [stats::mad] of each feature, possibly after centering.
```
mb706, line 35:
```suggestion
#'   If `robust` is `TRUE`, this is the median absolute deviation multiplied by 1.4826 (see [stats::mad]) of each feature, possibly after centering.
```

Issue #683, opened 2022-08-12
make single pipeop application easier


mb706:
```r
p = po("encode")
task_preproc = p$train(list(task))[[1]]
```

mb706:
potential problem: we don't want to tempt people into doing the wrong kind of preprocessing on train and test set

mb706:
suggestion (bb):
```r
outdata = preproc("encode", indata, paramlist, state = state)
```

mb706:
maybe we want to have `preproc(po("encode", paramlist))` instead, more consistent and would also work with PipeOps that are not in `po()`.

mb706:
could also overload the `%>>%` operator here.

mb706:
maybe: whenever state is given, or when optional var `predict = TRUE` (default FALSE) do prediction, otherwise training. We live with the fact that the state gets saved to the 1st arg pipeop. No reason not to make this work with graphs as well.

mb706:
```r
preproc(indata, graph, state = NULL, predict = !is.null(state))
```
Thoughts:
* we put 'indata' first so `|>` works well for transforming data.
* overloading `%>>%` is not so nice since `predict = ...` could not be set. We could have something like `%>>p%` for that, but that is probably too much clutter for not much gain so we currently choose not to do that.

some details:
* [ ] indata could also be a data.frame if we convert to a `Task` that does not have a target. See if TaskUnsupervised works (maybe it is too abstract and we need to subclass it here). Autotests for pipeops that do not use target at all should see if `preproc()` works for them for data.frames: `expect_datapreproc_pipeop_class` should take the `$data(cols = $feature_names)` and call call `preproc()` with it. Maybe expect_datapreproc_pipeop_class` needs an extra argument specifying whether target-less prepco() should be possible.
* [ ] would only work with single output graphs that accepts graphs; basically everything that works in a GraphLearner should also work here, so we could do similar checks as in the GraphLearner.
* [ ] Check if we can use assert_graph() on 2nd argument or if it breaks things, because 2nd argument should be modified by-reference (state should be set!).
* [ ] If no state is given and predict is FALSE (or not given, as it defaults to FALSE when state is NULL), graph$train() should be used and the graph should be modified by-reference.
* [ ] If state is given, the graph should be modified by-reference.
* [ ] If state is given and predict is FALSE, that is an error.
[...]

PR #882, opened 2025-02-23
New `preproc()` function for easier application of PipeOps or Graphs

advieser:
closes https://github.com/mlr-org/mlr3pipelines/issues/683

TODO:
- [ ] Discuss whether backends and other types are acceptable for `indata`
- [ ] check whether we need `assert_graph` (see issue)
- [ ] Discuss whether/how we want to check whether our graph can handle a targetless task (constructed from data.frame-like indata)
- [ ] Check whether we have to pay attention to anything when handling states
- [ ] Write `preproc.Graph`
- [ ] Write docs, think of good examples
- [ ] Think of test cases, especially how to test basic properties
- [ ] Modify `expect_datapreproc_pipeop_class`


mb706:
I think to be consistent we would use `as_graph(clone = FALSE)`, and check that every code path avoids cloning. This means extending [this](https://github.com/mlr-org/mlr3pipelines/blob/fb386de955adf1fd517af3d1423446a24672d08b/R/assert_graph.R#L118-L124), i.e. add an argument to `PipeOpLearner$initialize` `clone` (default `TRUE`) which would need to be set to `FALSE` here. Then we could use `preproc()` with a `Filter` as 2nd argument and the filter ends up being trained (I think?).

Using a `Learner` would not work because its output is not a `Task`, but still good to be consistent with respect to the `as_graph(clone = FALSE)` behaviour.

Please check if using a filter as 2nd argument would actually work, I think conceptually it would make sense.

advieser:
We decided against allowing Filters or Learners as `graph`, as
- Learners could not be modified in-place, since the state gets saved in the PipeOpLearner (resulting from call to `as_graph()`) and gets set to NULL on exit for the Learner itself.
- Filters could not be trained since when they get "converted" to a PipeOpFilter, when calling `as_graph()`, the setting of at least one of the hyperparameters is required.

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'
+#' @description
+#' Function that offers a simple and direct way to train or predict [`PipeOp`]s and [`Graph`]s on [`Task`][mlr3::Task]s,
+#' [`data.frame`]s or [`data.table`][data.table::data.table]s.
+#'
+#' Training happens if `predict` is set to `FALSE` and no `state` is passed to this function.
+#' Prediction happens if `predict` is set to `TRUE` and if the passed `Graph` or `PipeOp` is either trained or a `state`
+#' is explicitly passed to this function.
+#'
+#' The passed  `PipeOp` or `Graph` gets modified-by-reference.
```
mb706, line 11:
```suggestion
#' The passed  `PipeOp` or `Graph` gets modified by-reference.
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'
+#' @param indata ([`Task`][mlr3::Task] | [`data.frame`] | [`data.table`][data.table::data.table] )\cr
+#'   Data to be pre-processed.
+#' @param processor ([`Graph`] | [`PipeOp`])\cr
+#'   `Graph` or `PipeOp` which has exactly one input channel accepting a [`Task`][mlr3::Task] and one output channel.\cr
```
mb706, line 16:
```suggestion
#'   `Graph` or `PipeOp` accepting a [`Task`][mlr3::Task] and one output channel.\cr
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'   If `indata` is a [`data.frame`] or [`data.table`][data.table::data.table], the output channel must give a `Task` to be converted back into a
```
mb706, line 17:
```suggestion
#'   If `indata` is a [`data.frame`] or [`data.table`][data.table::data.table], the output channel must return a `Task` to be converted back into a
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'   `data.frame` or `data.table`. Additionally, `processor`s which only work on sub-classes of
+#'   [`TaskSupervised`][mlr3::TaskSupervised] will not be accepted, as it would be unclear which column was the `target`.\cr
```
mb706, line 19:
```suggestion
#'   [`TaskSupervised`][mlr3::TaskSupervised] will not accept [`data.frame`] or [`data.table`][data.table::data.table], as it would be unclear which column was the `target`.\cr
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'   Be aware that the `processor` gets modified-by-reference during training or if a `state` is passed to
+#'   this function. This especially means that the state of a trained `processor` will get overwritten.\cr
```
mb706, line 21:
```suggestion
#'   this function. This especially means that the state of a trained `processor` will get overwritten when `state` is passed.\cr
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'   You may want to use dictionary sugar functions to select a `processor` and to set its hyperparameters, e.g.
+#'   [`po()`][po] or [`ppl()`][ppl].
```
mb706, line 23:
```suggestion
#'   [`po()`] or [`ppl()`].
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#' @param state (named `list` | `NULL`)\cr
+#'   Optional state to be used for prediction, if the `processor` is untrained or if the current `state` of the `processor`
+#'   should be overwritten. Must be a complete and correct state for the respective `processor`.
```
mb706, line 26:
```suggestion
#'   should be overwritten. Must be a complete and correct state for the respective `processor`.
#'   Default `NULL` (do not overwrite `processor`'s `state`).
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#' @param predict (`logical(1)`)\cr
+#'   Whether to predict (`TRUE`) or train (`FALSE`). By default, this is `FALSE` if `state` is `NULL` (`state`'s default),
+#'   and `TRUE` otherwise.
+#' @return (`any` | [`data.frame`] | [`data.table`][data.table::data.table])\cr
+#' If `indata` is a `Task`, whatever is stored in the `processor`'s single output channel is returned.
```
mb706, line 31:
```suggestion
#' If `indata` is a `Task`, whatever is returned by the `processor`'s single output channel is returned.
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'   this function. This especially means that the state of a trained `processor` will get overwritten.\cr
+#'   You may want to use dictionary sugar functions to select a `processor` and to set its hyperparameters, e.g.
+#'   [`po()`][po] or [`ppl()`][ppl].
+#' @param state (named `list` | `NULL`)\cr
+#'   Optional state to be used for prediction, if the `processor` is untrained or if the current `state` of the `processor`
+#'   should be overwritten. Must be a complete and correct state for the respective `processor`.
+#' @param predict (`logical(1)`)\cr
+#'   Whether to predict (`TRUE`) or train (`FALSE`). By default, this is `FALSE` if `state` is `NULL` (`state`'s default),
+#'   and `TRUE` otherwise.
+#' @return (`any` | [`data.frame`] | [`data.table`][data.table::data.table])\cr
```
mb706, line 30:
```suggestion
#' @return `any` | [`data.frame`] | [`data.table`][data.table::data.table]:
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#' If `indata` is a `Task`, whatever is stored in the `processor`'s single output channel is returned.
+#' If `indata` is a [`data.frame`] or [`data.table`][data.table::data.table], an object of the same class is returned, or
+#' if the `processor`'s output channel does not return a `Task`, an error is thrown.
+#'
+#' @section Internals:
+#' If `processor` is a [`PipeOp`], the S3 method `preproc.PipeOp` gets called first, converting the `PipeOp` into a
+#' [`Graph`] and wrapping the `state` appropriately, before calling the S3 method `preproc.Graph` with the modified objects.
+#'
+#' If `indata` is a [`data.frame`] or [`data.table`][data.table::data.table], it is used to construct a
+#' [`TaskUnsupervised`][mlr3::TaskUnsupervised] internally. This implies that `processor`s which only work on sub-classes
```
mb706, line 40:
```suggestion
#' If `indata` is a [`data.frame`] or [`data.table`][data.table::data.table], a
#' [`TaskUnsupervised`][mlr3::TaskUnsupervised] is constructed internally. This implies that `processor`s which only work on sub-classes
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#'
+#' @export
+#' @examples
+#' library(mlr3)
+#' task = tsk("iris")
+#' pop = po("pca")
+#'
+#' # Training
+#' preproc(task, pop)
+#' # Note, that the PipeOp gets trained through this
```
mb706, line 51:
```suggestion
#' # Note that the PipeOp gets trained through this
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#' pop$state$sdev
+#'
+#' # Piping multiple preproc() calls, using dictionary sugar to set parameters
+#' outdata = preproc(tsk("penguins"), po("imputemode", affect_columns = selector_name("sex"))) |>
+#'   preproc(po("imputemean"))
+#' outdata$missings()
+#'
+#' # Use preproc with a Graph
+#' gr = po("pca", rank. = 4) %>>% po("learner", learner = lrn("classif.rpart"))
+#' preproc(tsk("sonar"), gr) # returns NULL because of the learner
```
mb706, line 76:
```suggestion
#' preproc(tsk("sonar"), gr)  # returns NULL because of the learner
```

R/preproc.R
```diff
@@ -0,0 +1,149 @@
...
+#' preproc(tsk("sonar"), gr, predict = TRUE)
+#'
+#' # Training with a data.table input
+#' # Note that this treats the target column as if it were a feature ...
```
mb706, line 80:
```suggestion
#' # Note that `$data()` drops the information that "Species" is the target.
#' # It gets handled like an ordinary feature here.
```

R/preproc.R
```diff
@@ -0,0 +1,151 @@
...
+  preproc(indata, processor, state, predict)
+}
+
+#' @export
+preproc.Graph = function(indata, processor, state = NULL, predict = !every(state, is.null)) {
+  assert_flag(predict)
+  if (nrow(processor$output) != 1) {
+    stop("'processor' must have exactly one output channel.")
+  }
+  # Note: We also expect the processor to only have a single input channel which only accepts Tasks. However, we put the
```
mb706, line 116:
```suggestion
  # Note: We do accept graphs with more than one input; indata is sent to all of them,
  # making use of Graph$train's single_input = TRUE.
```

R/preproc.R
```diff
@@ -0,0 +1,151 @@
...
+  #       burden of checking against this on Graph's check_types.
```
mb706, line 117:
```suggestion
```

mb706:
Todo:
* test multi-input graph
* test multiplicityexply
* test non-task output error with debug-pipeop

inst/testthat/helper_functions.R
```diff
@@ -468,6 +470,16 @@ expect_datapreproc_pipeop_class = function(poclass, constargs = list(), task,
...
   }
+
+  # Test that preproc() works with PipeOp
+  expect_no_error(preproc(task, po))
+  expect_no_error(preproc(task, po, predict = TRUE))
+  if (po$input$train == "Task") {  # implies that PipeOp does not require target column
+    dtout = expect_no_error(preproc(task$data(cols = task$feature_names), po))
+    expect_data_table(dt_out)
+    dtout = expect_no_error(preproc(task$data(cols = task$feature_names), po, predict = TRUE))
+    expect_data_table(dt_out)
```
mb706, line 481:
```suggestion
  expect_no_error(mlr3pipelines::preproc(task, po))
  expect_no_error(mlr3pipelines::preproc(task, po, predict = TRUE))
  if (po$input$train == "Task") {  # implies that PipeOp does not require target column
    dtout = expect_no_error(mlr3pipelines::preproc(task$data(cols = task$feature_names), po))
    expect_data_table(dtout)
    dtout = expect_no_error(mlr3pipelines::preproc(task$data(cols = task$feature_names), po, predict = TRUE))
    expect_data_table(dtout)
```

Issue #881, opened 2025-02-23
Obscure error in `PipeOpADAS` due to unobserved factor levels


advieser:
```R
library(mlr3pipelines)
library(mlr3)
op = po("adas")
task = tsk("iris")$filter(c(1:20, 40:89))
op$train(list(task))

# Error in if (as.numeric(num_syn_i[i]) > 0) { :  
# missing value where TRUE/FALSE needed  
# This happened in PipeOp adas's $train()  
# Additionally: Warning message:  
# In get.knnx(data, query, k, algorithm) : k should be less than sample size!  
# This happened in PipeOp adas's $train()
```

The error occurs because there is an unobserved factor level in the target column, i.e.:
```R
table(task$truth())
#    setosa versicolor  virginica 
#        50         50          0 
```
Internally, in `smotefamily::adas()`, this leads to `P_size` being `0` and `1:P_size` evaluating to `c(1, 0)`. This leads to calling `numeric(0) > 0`.
```R
for(i in 1:sizeP) {    
  if(as.numeric(num_syn_i[i])>0)
  ...
}
```
The warning also results from `P_size = nrow(P_set) = 0` since `FNN` calculates the sample size as `nrow(P_set)`.

Possible solution:
We might remove unobserved factor levels from target internally, and add them back in again? Don't know whether this makes sense in the context of upsampling. Otherwise, we may just want to check for this and throw a more informative error.

advieser:
There is also this in `PipeOpSmote`:
```R
op = po("smote")
task = tsk("iris")$filter(c(1:100))
op$train(list(task))
# Error in runif(sum_dup) : invalid arguments  
# This happened in PipeOp smote's $train()  
# Additionally: Warning message:  
# In get.knnx(data, query, k, algorithm) : k should be less than sample size!  
# This happened in PipeOp smote's $train()
```
Here, `sum_dup` is `Inf` because in `smotefamily::n_dup_max()` a divide-by-zero occurs, again since `P_size` is `0`.

mb706:
I think we should be robust for these kinds of cases. Maybe also check if we can add this to autotests for preproc pipeops that work with target?

advieser:
(probably also check `PipeOpBLSmote`)

advieser:
```r
op = po("adas")
task = tsk("iris")
op$train(list(task))
# Error in if (as.numeric(num_syn_i[i]) > 0) { :  
# missing value where TRUE/FALSE needed  
[...]

PR #885, opened 2025-02-28
Fixes implicit dropping of unseen target levels in SMOTE-like PipeOps

advieser:
closes https://github.com/mlr-org/mlr3pipelines/issues/881


advieser:
Concerning the empty target levels, we decided to also drop them in PipeOps from the `themis` package as well to have consistent behavior, i.e. dropping empty target levels instead throwing an error (which `themis` would do itself otherwise).

PR #897, opened 2025-03-28
Add error message for predicting with untrained PipeOp or Graph

### LINKED ISSUE #893: $predict should throw an error if PO is not trained
Otherwise we get weird error messages when `state` is unexpectedly NULL.

Weird that we did not have this already.... maybe check if we do have it somewhere else and it does not apply to all pipeops, for some reason? (e.g. we only check in Graph?).

Example:
```r
po("imputeoor")$predict(list(tsk("iris")))
#> Error: Input task during prediction of imputeoor does not match input task during training.
```

may be a minor breaking change if someone unsed "stateless" POs without training them, before. But this should really not have been allowed to begin with.

advieser:
closes #893

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,5 @@
 # mlr3pipelines 0.7.2-9000
-
+- Added missing error for predicting with untrained `PipeOp`s / `Graph`s.
 
 # mlr3pipelines 0.7.2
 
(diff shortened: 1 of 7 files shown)


advieser:
Added error message for Graph would not be strictly necessary, I think, because PipeOps and Learners would throw error. However, error message makes more sense this way. 

I have not added something similar to `GraphLearner`, as it inherits from `Learner` which already has an appropriate error message. Only slight discrepancy is that the error message says Learner, and not GraphLearner, in that case.

PR #898, opened 2025-03-29
refactor: use more `ps$set_values()`

m-muecke:
@mb706 check if you want/like this, this standardises to use more `ps$set_values()` like in the other mlr3 packages


mb706:
conflicted about this; the point of `$values = list()` is that the previous state of `$values` is overwritten. `$set_values()` does not change values that are already set; even if it does the same thing, the information that there is no relevant previous state is lost. In reality, we should probably use the `init = ` functionality of the new paradox syntax. Will leave this open for now and think about it

Issue #910, opened 2025-05-16
Fail on cran because of missing kknn


mb706:
https://cran.r-project.org/web/checks/check_results_mlr3pipelines.html

The example should probably be made to run optionally if kknn. Besides that, maybe we can remove the kknn dependency altogether, since it is currently [archived on CRAN](https://cran.r-project.org/web/packages/kknn/index.html) and we don't know whether it will be coming back.

PR #911, opened 2025-05-16
Remove `kknn` dependency

advieser:
Since we removed `LearnerClassifKKNN` from mlr3learners, see https://github.com/mlr-org/mlr3learners/pull/339, because `kknn` was archived on CRAN.
closes https://github.com/mlr-org/mlr3pipelines/issues/910


mb706:
merging this to go ahead with mlr3 weights update revdep checks

Issue #912, opened 2025-05-17
Compatibility with mlr3 "weights" update


mb706:
After [this](https://github.com/mlr-org/mlr3/pull/1124) was merged, we now need to handle the new and different weights column types

PR #914, opened 2025-05-19
Fixes to achieve compatibility with mlr3 weights update

advieser:
closes https://github.com/mlr-org/mlr3pipelines/issues/912
- GraphLearner tests: `run_experiment` now checks behavior with weights and we had contradictory `$properties` (without `"weights"`) and `$use_weights` set to `"use"` since we modify the properties manually as some workaround (didn't check whether we still need that workaround).
- `ppl("robustify")` tests: Removed overwriting of `lrn$properties` that (1) deleted `"weights"` property necessary when calling `$.use_weights()` internally and (2) was seemingly illogical, overwriting the properties to `c("multiclass", "twoclass")` when that Learner was to be trained on a regression task.
- `LearnerRegrAvg` / `LearnerClassifAvg`: Now uses `$set_values()` to modify the ParamSet of OptimizerNloptr (which is used by default) to avoid overwriting the default value of its new hyperparameter `approximate_eval_grad_f` (https://github.com/mlr-org/bbotk/commit/9519f956a0a12b9a26644a74c1795eccaa005a5b).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * Added missing error for predicting with untrained `PipeOp`s / `Graph`s.
 * Fix: Corrected typo in the hyperparameter name `use_parallel` of `PipeOpVtreat`.
+* Fix: New hyperparameter of `bbotk::OptimizerBatchNLoptr` does not get overwritten in `LearnerClassifAvg` / `LearnerRegrAvg`'s internal `optimize_weights_learneravg` function.
 
 # mlr3pipelines 0.7.2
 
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -2,6 +2,7 @@
 
 * Added missing error for predicting with untrained `PipeOp`s / `Graph`s.
 * Fix: Corrected typo in the hyperparameter name `use_parallel` of `PipeOpVtreat`.
+* Fix: New hyperparameter of `bbotk::OptimizerBatchNLoptr` does not get overwritten in `LearnerClassifAvg` / `LearnerRegrAvg`'s internal `optimize_weights_learneravg` function.
```
mb706, line 5:
```suggestion
* Fix: Do not overwrite initial hyperparameter settings of `bbotk::OptimizerBatchNLoptr` in `LearnerClassifAvg` / `LearnerRegrAvg`'s internal `optimize_weights_learneravg` function.
```

Issue #918, opened 2025-06-16
Remove old `paradox` compatability


advieser:
See https://github.com/mlr-org/mlr3pipelines/blob/cd9b3e1ec2ac322c412d366d6ca59a145e43ad97/R/zzz.R#L33 and places where this is used.
paradox version number in DESCRIPTION should then be increased.

PR #931, opened 2025-07-24
Removes compatibility for old `paradox`

advieser:
Closes https://github.com/mlr-org/mlr3pipelines/issues/918


advieser:
I'm not sure whether the paradox version number is now correct, since I could not tell by paradox's NEWS or history of `ParamSet.R` when exactly these changes were made. I now set it to 1.0.0 to be safe.

PR #932, opened 2025-07-24
Fix `preproc()` examples

advieser:


advieser:
We have to remove `NMF` from `packages` in `PipeOpNMF`'s `super$initialize()` as it would otherwise get loaded through `require_namespaces(self$packages)` in `PipeOp`'s `train` and `predict` methods, and we then couldn't deduce what the search path looked like before (from within `PipeOpNMF` at least).

advieser:
Current problems:
- Need a better/more elegant solution for tests because of the namespace loading and attaching.
- The current solution does not work for predicting if the PipeOp was not trained before. This is because the `state` gets accessed from `PipeOp$predict()` before our solution is run in the inheriting `PipeOpNMF`'s `$.predict_dt()`. If the `NMF` namespace was not loaded before the state is accessed, S4 methods dispatch looks for the S4 class `NMFfit` of the object stored in the state. In the course of this lookup, the `NMF` package gets loaded. So then we reach our solution in the `$predict_dt()` call, the search path looks as if `NMF` was loaded before calling the PipeOp's predict method.
We could say, that we live with this as it should be rare enough that the PipeOp's predict is called without it having been trained first. For this to happen, without the respective NMF functions to have been called first by the user (and modifying the search path at places where we couldn't do anything anyways), the state would have to be read from disk.

PR #934, opened 2025-07-28
PipeOpImpute predict-imputes factors when empty_level_control == "never"

mb706:


mb706:
todo:
* test robustify pipeline removes all NAs when unseen factors appear
* test impute sample, impute learner for the same case

PR #987, opened 2026-03-09
Fix all.equal method registration

MichaelChirico:
This is due to https://github.com/r-lib/roxygen2/issues/1587

After next {roxygen2} release this will "just go away", AIUI, so you could also just wait


mb706:
Thanks!

PR #998, opened 2026-06-07
Fix: NMF rejects features with NA or Inf

### LINKED ISSUE #983: `PipeOpNMF`'s `.select_cols` does not handle `NA`s gracefully
```R
library(mlr3)
library(mlr3pipelines)
op = po("nmf")
op$train(list(tsk("penguins")))
# Error in `.__Task__data()`:
# ! Assertion on 'cols' failed: Must be a subset of {'..row_id','bill_depth','bill_length','body_mass','flipper_length','island','sex','species','year'}, but has additional elements {'NA'}.
# This happened in PipeOp nmf's $train()
# Hide Traceback
#      ▆
#   1. ├─op$train(list(tsk("penguins")))
#   2. │ └─mlr3pipelines:::.__PipeOp__train(...)
#   3. │   ├─base::withCallingHandlers(...) at mlr3pipelines/R/PipeOp.R:315:7
#   4. │   └─private$.train(input) at mlr3pipelines/R/PipeOp.R:316:9
#   5. │     └─mlr3pipelines:::.__PipeOpTaskPreproc__.train(...)
#   6. │       └─private$.train_task(intask) at mlr3pipelines/R/PipeOpTaskPreproc.R:211:7
#   7. │         └─mlr3pipelines:::.__PipeOpTaskPreproc__.train_task(...)
#   8. │           └─task$data(cols = cols) at mlr3pipelines/R/PipeOpTaskPreproc.R:267:7
#   9. │             └─mlr3:::.__Task__data(...)
#  10. │               └─checkmate::assert_subset(cols, self$col_info$id)
#  11. │                 └─checkmate::makeAssertion(x, res, .var.name, add)
#  12. │                   └─checkmate:::mstop(...)
#  13. │                     └─base::stop(simpleError(sprintf(msg, ...), call.))
#  14. └─mlr3pipelines (local) `<fn>`(`<smplErrr>`)

# mlr_pipeops$add("nmf", PipeOpNMF)
```

The source of the error lies here:
https://github.com/mlr-org/mlr3pipelines/blob/0ff0e8f8066c89a90a0fb29ab5525f5be10b9d43/R/PipeOpNMF.R#L206-L211
Here, if a feature column contains `NA`s, `all(x >= 0)` will return an NA. This is then part of the return vector, which should be a vector of column names.

The solution to this depends on whether `NMF::nmf()` can handle `NA`s. If no feature fits our criterion (non-negative and potentially fully observed), we probably don't want an error and instead just continue with the PipeOp not doing anything? It would be good to add an early exit in the train and predict methods in that case.

advieser:
closes https://github.com/mlr-org/mlr3pipelines/issues/983


mb706:
Long term, we do not want to silently ignore features with negative values, we want the user to make the deliberate choice (and get an error message if there is unexpected behaviour). So long term we we want an error if negative cols are selected, but we want to ease into this, giving a warning so far.

Planned behaviour:
1. give an error if missing values are present (as is done here)
2. introduce a selector for non-negative numeric / integer features
3. PipeOpNMF should give a warning if non-negative features are present but still drop them (as is done already). The warning should indicate that this will be an error in the future, and suggest to use the non-negative `Selector`.
4. Make sure the warning is not given if the selector (non-negative selector, or e.g. a selector by name) is present that drops the cols with negative values.

(5. at some point in the future, turn the warning into an error)

advieser:
Maybe add:
- selector_positive, selector_negative, selector_non_negative, selector_non_positive, selector_non_zero
- selector_non_missing (while we are at it)

advieser:
TODO: 
- Change keep_na default to TRUE (and change call in NMF appropriately).
- Add comment that affect_columns is run before .select_cols().

PR #1000, opened 2026-06-09
skip isomap tests/example when igraph is unavailable

be-marc:
dimRed's Isomap method requires igraph at runtime. When igraph is not installed, the PipeOpIsomap example and tests fail because they only guarded against dimRed/RSpectra/stats. Add igraph to the guards. This happens every time when I run R CMD check on mlr3pipelines because the installation of igraph fails on my revdep machine.


advieser:
TODO:
- News
- Add RANN to checks also
- Add these packages to `packages` argument in `initialize`.

@be-marc: I'll do these and then merge this. Thanks!

PR #1001, opened 2026-06-09
refactor: use more stopf()

m-muecke:


advieser:
Hey Max, FYI, we specifically wanted `call. = FALSE` here. I plan to make a PR to mlr3misc so that`stopf` accepts `call.` and passes it on. We then wanted to refactor this occurance.

PR #1006, opened 2026-06-20
feat(targetinvert): propagate the prediction extra slot

m-muecke:
The extra slot is currently dropped, causing an issue for extension packages that rely on the extra slot provided in mlr3. Since it's a target trafo I would argue that the no-op for the extra slot is a simple decision.


m-muecke:
I actually don't need this anymore, so your call to merge it or not.

advieser:
OK, I'll talk with Martin and we'll see.

PR #1008, opened 2026-07-07
fix: as_learner() forwards dots to the GraphLearner constructor

m-muecke:


mb706:
currently fails because GraphLearner does not have `discard_state`.
* Modify `as_learner.Graph` to have the named arguments that GraphLearner$initialize offers; in the form `as_learner.Graph(x, clone = TRUE, ..., id = [whatever], param_vals = list(), [...])`; i.e. named arguments after `...`.
* support `discard_state` at the level of `as_learner.Graph` by just setting `state = NULL` if `clone && discard_state`; this might in very few cases produce unneeded clones, but is much simpler.

@advieser

advieser:
If we explicitly support `discard_state`, do we actually still need the constructor arguments as named arguments?

PR #1013, opened 2026-07-23
feat: add boundary date features to PipeOpDateFeatures

m-muecke:
@mb706 not a big fan of enabling all the features per default, but it does follow fastapi implementation. Setting them to false would also be backwards compatible.


advieser:
We decided to initialize them as `FALSE` for backwards compatibility.

PR #1014, opened 2026-07-24
test: replace pima task usage after mlbench dataset removal

be-marc:
The `pima` task is being removed from mlr3 because the `PimaIndiansDiabetes2` dataset was removed from the mlbench package. This PR removes replaces it with `tsk("diabetes")`.


advieser:
We decide to export our own diabetes task (identical to the one in mlr3) for a couple patches to be compatible to older mlr3 versions. 

A consequence of this, albeit a negligible one, is that our task's man page will not be able to rely on roxygen2 templates used in mlr3. 

Added tests to check that the mlr3 version and our version are identical. These are skipped if `load_task_diabetes` does not exist in the `mlr3` namespace, i.e. for versions of mlr3 that have not yet added the new task.

advieser:
The compatibility shim will need to be updated to fit https://github.com/mlr-org/mlr3/pull/1497.

R/TaskClassif_diabetes.R
```diff
@@ -0,0 +1,105 @@
...
+  b = as_data_backend(make_diabetes_fixture())
+  task = TaskClassif$new(id, b, target = "diabetes", positive = "pos", label = "Synthetic Diabetes")
+  b$hash = "mlr3::mlr_tasks_diabetes"
+  task$man = "mlr3pipelines::mlr_tasks_diabetes"
+  task
+}
+
+supply_diabetes = function() {
+  if (tsk()$has("diabetes")) return(invisible(NULL))
+  tsk()$add("diabetes", load_task_diabetes)
```
mb706:
```suggestion
  tsk()$add("diabetes", function(id) tsk("pima", id = id))
```

advieser:
We assume that a user either uses old mlr3 and old mlbench, not old mlr3 with up-to-date mlbench.

PR #1017, opened 2026-08-06
Feat/best valid scores

sebffischer:


DESCRIPTION
```diff
@@ -110,6 +110,8 @@ Suggests:
     dimRed,
     RSpectra,
     RANN
+Remotes:
+    mlr-org/mlr3
```
mb706:
```suggestion
```

mb706:
@advieser if tests fail because of old version, please add a version guarded skip to the new tests here

Issue #1018, opened 2026-08-14
`preproc` examples use pipe `|>` introduced with newer `R` version


advieser:
https://github.com/mlr-org/mlr3pipelines/blob/6742a41f32cbde8787cb481a281ef72eae35acf6/R/preproc.R#L71-L74

We currently depend on R version >=3.3.0, but this was introduced only later with 4.1.0.

PR #1026, opened 2026-09-05
Fix CRAN compatibility issues

mb706:


advieser:
Closes https://github.com/mlr-org/mlr3pipelines/issues/1018

